PR #14, opened 2015-05-28
newName property in SampleApp class in unused

jgolla:
The newName property is the SampleApp class is not used. It's a little confusing since the parameter to addnew is named "newname" as well.


mgechev:
Thanks for the catch! :-)

jgolla:
No problem. Thanks for writing this. I have been playing with it for a couple of days now and just noticed that.

PR #23, opened 2015-06-25
Simplify About component code

robertpenner:
Used TypeScript shorthand.
Removed names property and referenced the list in the template.


mgechev:
Thanks for the catch!  :-)

PR #27, opened 2015-06-29
Update Angular typings to v2.0.0-alpha28

dstockhammer:


mgechev:
Thanks for the PR!

ludohenin:
Just one comment on that. `Http` is defined in `angular2/angular2` where it should actually be in `angular/http` module instead, not a big deal at all but for you to know (if you don't)

mgechev:
@ludohenin thanks for pointing this out, didn't notice it.

PR #34, opened 2015-07-15
Update to Angular 2.0.0-alpha.31. Also adding nice example usage of Http class.

NathanWalker:
IMO, An adequate `seed` repo should at least contain example usages of some primary classes like Http.

### DIFF
--- a/app/app.css
+++ b/app/app.css
@@ -1,3 +1,12 @@
+h1,
+h2,
+h3,
+h4,
+h5 {
+  padding:0;
+  margin:0;
+}
+
 .sample-app-content {
   font-family: Verdana;
 }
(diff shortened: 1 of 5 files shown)


ludohenin:
@NathanWalker I don't think this will be accepted. Example usage are left to [AngularShowcase](https://github.com/AngularShowcase)

mgechev:
@NathanWalker thanks for the PR! We're trying to keep the seed project as minimalistic as possible and provide all extra examples in [AngularShowcase](https://github.com/AngularShowcase/).

@ludohenin added HTTP example in the [sample application](https://github.com/AngularShowcase/angular2-sample-app) in [AngularShowcase](https://github.com/AngularShowcase/), you can take a look.

NathanWalker:
That makes complete sense, I'll keep that in mind in future!

PR #37, opened 2015-07-23
feat: update to angular 2.0.0-alpha.32. Also simplified typings

NathanWalker:
Simplified typings to be consistent approach across most all seed projects now in managing custom typings vs. tsd installed definitions. Suggestion by @gdi2290


gdi2290:
:+1:

app/index.html
```diff
@@ -3,8 +3,10 @@
 <head>
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
-  <base href="<%= APP_BASE %>">
   <title>My Angular 2 App</title>
+  <base href="<%= APP_BASE %>">
+  <!-- ignore: Aplha 32 router fix -->
+  <script>baseElement = document.querySelector('base');baseElement.attr = baseElement.getAttribute;</script>
```
NathanWalker: This is also what @gdi2290 did as well:
https://github.com/angular-class/angular2-webpack-starter/blob/0122d866cfcffb76150aa3047e6f9adb1959733d/src/public/index.html#L15-16

NathanWalker:
Additionally I issued a PR to DefinitelyTyped for systemjs based on what you guys had mocked out:
https://github.com/borisyankov/DefinitelyTyped/pull/5043
:)

ludohenin:
@NathanWalker I have a lot of typing error when building the app. Do you also have some ?

ludohenin:
A lot of `Duplicate identifier` actually and few `not found` ones.

ludohenin:
@NathanWalker Found the duplicates, this one `/tsd_typings/custom.system.d.ts` is to be removed.

NathanWalker:
Probably best to remove the `typings` folder in root altogether if one was still there.
Then `tsd reinstall` to get them fresh.

Custom type definitions are to go `only` in `app/typings`... everything in `tsd_typings` should be straight from the `tsd install/reinstall` command based off `tsd.json` :+1:

ludohenin:
right

NathanWalker:
I'll clean that up... 1 min, I'll commit a fix.

ludohenin:
but the compiler still can't find abs ref `<reference path="/app/typings/_custom.d.ts" />` in app.ts, home.ts...

NathanWalker:
The above commit https://github.com/NathanWalker/angular2-seed/commit/eb1e3d1c2d7bd7655602c0e9e4b869fb3fe0ce36 should fix it.

ludohenin:
Looks good :-)
`experimentalDecorators` set to `true` in `tsconfig.js` and that should be perfect.
thanks,

NathanWalker:
Ah yes, good catch! Thanks @ludohenin I got the absolute references cleaned up too, all is solid :+1:

ludohenin:
@mgechev I think it is ready to merge

mgechev:
Yep, just too a look. Looks great!

PR #38, opened 2015-07-24
Remove unused ngFor class

redian:
Remove unused ngFor class


mgechev:
Thanks for the catch!

Issue #41, opened 2015-07-31
Allow live reloading in browser


mmrath:
It would be nice to have auto reloading in browser when files are saved.

mgechev:
Yep, makes sense. I'll take this one.

PR #42, opened 2015-07-31
Add livereload for prod and dev

mgechev:
Fix for #41 
@ludohenin would you take a look and merge if you think its fine.


ludohenin:
@mgechev looks great

PR #53, opened 2015-08-19
Update to angular2-alpha34

mgechev:
There's a workaround with System.js in order to make it work. It shouldn't be necessary when alpha35 is relased.


mgechev:
I'll merge it and update to alpha35 today.

PR #60, opened 2015-08-27
Gulp src reference

aboeglin:
I have extended the Configuration variables and used them to define the PATH. Like this we can just do changes to the paths at one point. So we can now change the source and destination paths easily.

I have also added a folder created by webstorm in the .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,9 @@ node_modules
 # Users Environment Variables
 .lock-wscript
 
+#IntelliJ configuration files
+.idea
+
 dist
 dev
 lib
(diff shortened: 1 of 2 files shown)


ludohenin:
In my opinion, this needs some clean-up.
* intendation: 2 spaces
* use `join(var, 'tesxt')` instead of ` var+'text'` (at least `var + 'text')

mgechev:
Yeah, I agree with @ludohenin. @aboeglin would you fix these two points. Otherwise it is ready to be merged! Thanks for the PR!

aboeglin:
Agreed, my editor indented to 4 spaces, I'll fix it and use join() where appropriate. I'll try to do it tonight.

aboeglin:
Also now that I think about it, I could also update the readme file in the Configuration section to briefly describe it.

aboeglin:
Ok it should be good now !

mgechev:
Looks good to me! Thanks for the PR!

PR #63, opened 2015-08-28
Fixed wrong use of path.join() in gulpfile clean.app.dev task

mjwwit:
path.join() was being used to concatenate a gulp excluding `'!'` with a path, resulting in the PATH.dest.dev.lib folder being deleted when serving dev.
This fix removes the wrong path.join() in favor of simple string concatenation.


ludohenin:
@mjwwit Thanks for the PR. Will merge it.

PR #87, opened 2015-09-22
Making the about form example work as expected. Using Enter inside an…

NathanWalker:
Hey guys, since the first time I encountered this seed project, I felt this was needed.
I made the about form example work as expected. Using `Enter` inside an input to submit a form meets baseline usability expectations. The Add button works as well.

I think people should be seeded with a setup that at least meets basic usability expectations. I thought if this would be considered something for the showcase app, but I think this is a fundamental expectation :+1:

### DIFF
--- a/app/components/about/about.html
+++ b/app/components/about/about.html
@@ -5,7 +5,7 @@
 <p>
   You want more? Add them yourself!
 </p>
-<form (submit)="addName(newname); $event.preventDefault();">
+<form (submit)="addName(newname)">
   <input #newname>
   <button type="submit">Add</button>
 </form>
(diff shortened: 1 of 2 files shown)


app/components/about/about.html
```diff
@@ -5,10 +5,10 @@
 <p>
   You want more? Add them yourself!
 </p>
-<p>
+<form (submit)="addName(newname); $event.preventDefault();">
```
ludohenin: I think `return false;` on the addName method should prevent submit default behaviour.
That would look a little bit prettier.

NathanWalker: Agreed. Thanks!

NathanWalker:
Thanks @ludohenin, I updated the PR, added a comment in the `addName` method as well as modified the return `type`.

mgechev:
Makes sense. Merging.

PR #91, opened 2015-09-25
fix(build): make it compatible with npm3

mgechev:
Two fixes:
- `require.resolve` cant find module `traceur` when npm 3.3.3 is used
- move from `tiny-lr` to `mini-lr`

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -32,7 +32,7 @@ var express = require('express');
 var serveStatic = require('serve-static');
 var openResource = require('open');
 
-var tinylr = require('tiny-lr')();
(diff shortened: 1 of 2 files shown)


elwayman02:
:+1:

elwayman02:
@mgechev fyi, the tiny-lr project was updated and I've been added as a maintainer, so I'm going to deprecate mini-lr in the meantime to keep the community from fracturing while hopefully giving the original project better support. Please move back to tiny-lr when you are ready to upgrade your dependencies, and sorry for the back-and-forth!

mgechev:
Thanks for that!

PR #92, opened 2015-09-25
Problem with slashes after using require.resolve

svrcekmichal:
1. On windows, there's problem with require.resolve, which resolve to path with `\`. Gulp is not able to split it by `/`, and therefore unable to inject lib scripts.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -24,6 +24,7 @@ var del = require('del');
 var fs = require('fs');
 var path = require('path');
 var join = path.join;
+var slash = require('slash');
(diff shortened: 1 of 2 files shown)


mgechev:
Thanks for the PR!

ludohenin:
I'm very surprised as this has been working on Win 7 machine from the beginning.

svrcekmichal:
@ludohenin It was working because paths to files was hardcoded with `/` and injected that way, but after change to `require.resolve` the result is windows path with `\` where `.split('/')` makes no sense :)

PR #95, opened 2015-09-27
divide gulp tasks

jerryorta-dev:
This pr is something to consider for scalability. While it may be overkill in the current state of this seed project, it would be helpful when developers use this seed and start to add their own workflow.


mgechev:
Hey @jerryorta-dev. Looks great to me! Would you change the indentation to 2 spaces in order to be consistent?

//cc @ludohenin what are your thoughts?

ludohenin:
yeah, looks great!

PR #104, opened 2015-10-03
Adding gulp-less plugin task

abhishek-agashe:


mgechev:
Hey @abhishek-agashe it looks cool but we're trying to keep the seed project as minimalistic as possible, we don't want to make it opinionated about technologies to be used inside the project.

PR #107, opened 2015-10-03
Added missing semicolon

tarlepp:


ludohenin:
Yeah. That why I think we need a linter :)

tarlepp:
Yeah, I'll try to make gulp task for that.

Issue #102, opened 2015-10-02
Add tslint to the build workflow


ludohenin:

tarlepp:
Hmm, this one is solved - right?

ludohenin:
yep, you're right.

ludohenin:
Solved here #109

PR #109, opened 2015-10-03
Added js and ts linter tasks

tarlepp:


tarlepp:
Related to #107

tarlepp:
This solves #102

ludohenin:
Thanks @tarlepp, my first comments:
* I'm not fan of the option `sort-object-literal-keys` (looks better as it was before)
* for file names I'd prefer something like `tslint.dev.js` & `jshint.dev.js`
* not sure if `build.assets.dev` task is the appropriate one to lint the code (will check that)

//cc @mgechev any comments?

tarlepp:
I'll wait comments from @mgechev and after that I'll make necessary changes to PR.

mgechev:
I see the linting as task part of the `build.dev` task.

I also think, we should open a separate issue for discussions around the rules inside `tslint.json`

tarlepp:
rgr, I'll make necessary changes to this PR asap

tarlepp:
Okay, now I think all is done as you like.

tslint.json
```diff
@@ -0,0 +1,63 @@
+{
+  "rules": {
+    "class-name": true,
+    "comment-format": [true, "check-lowercase", "check-space"],
```
mgechev: Not sure if `check-lowercase` is not too restrictive.
What `true` stands for?

tslint.json
```diff
@@ -0,0 +1,63 @@
...
+    "curly": true,
+    "eofline": true,
+    "forin": true,
```
mgechev: There's a practice of omitting the `if` statement because of [performance reasons](https://jsperf.com/hasownproperty-vs-in-vs-undefined/12). I'm not sure if we're not getting to opinionated with this one.

tslint.json
```diff
@@ -0,0 +1,63 @@
...
+    "label-position": true,
+    "label-undefined": true,
+    "max-line-length": [true, 140],
+    "member-ordering": [true,
+      "public-before-private",
+      "static-before-instance",
+      "variables-before-functions"
+    ],
+    "no-arg": true,
+    "no-bitwise": true,
```
mgechev: Not completely sure on this either.

tslint.json
```diff
@@ -0,0 +1,63 @@
...
+    "no-console": [true,
```
mgechev: Why would we want to forbid access to these methods?

tslint.json
```diff
@@ -0,0 +1,63 @@
...
+      "debug",
+      "info",
+      "time",
+      "timeEnd",
+      "trace"
+    ],
+    "no-construct": true,
+    "no-debugger": true,
```
mgechev: Same here, this is dev environment so access to `debugger` makes sense.
May be we can create two separate `tslint` config files when we create a production build (once ng2 is officially released).

tslint.json
```diff
@@ -0,0 +1,63 @@
...
+    "no-use-before-declare": true,
+    "no-var-keyword": true,
+    "one-line": [true,
+      "check-open-brace",
+      "check-catch",
+      "check-else",
+      "check-whitespace"
+    ],
+    "quotemark": [true, "single"],
+    "radix": true,
```
mgechev: Not sure if this is necessary given the target browsers will be IE9 and newer.

tslint.json
```diff
@@ -0,0 +1,63 @@
...
+    "semicolon": true,
+    "sort-object-literal-keys": false,
+    "triple-equals": [true, "allow-null-check"],
```
mgechev: What `allow-null-check` does?

tarlepp:
Hmm, do I make those changes on this PR or do we continue this on #113 ?

tslint.json
ludohenin: it allows `myVar == null` but not `myVar == somevalue` (`myVar === someValue` would be expected)

mgechev:
Yeah, lets wait a bit for comments by @ludohenin and @NathanWalker. You can commit here, squash the commits to a single one and I'll merge.

tarlepp:
I changed ```tslint.json``` to match @mgechev proposal from #113

NathanWalker:
I am slammed this week and haven't had a chance to review fully but I agree
with @mgechev proposal from #113, as long as not too restrictive. +1
Thanks for the contribution!

mgechev:
@tarlepp awesome! I'm merging your PR!

PR #111, opened 2015-10-03
component specific css example

abhishek-agashe:
Adding related CSS inside of component folder, that way all assets required by component can stay at same level.


mgechev:
Thanks for the PR

PR #117, opened 2015-10-05
Add system-polyfills file to the lib/ dist directory

ryzy:
system-polyfills.js might be included by some browsers (i.e. IE) and SystemJS tries to load it from lib/ directory.

### DIFF
--- a/tools/workflow.config.js
+++ b/tools/workflow.config.js
@@ -34,6 +34,7 @@ var PATH = {
       require.resolve('reflect-metadata/Reflect.js'),
       require.resolve('reflect-metadata/Reflect.js.map'),
       require.resolve('systemjs/dist/system.src.js'),
+      require.resolve('systemjs/dist/system-polyfills.js'),
       APP_SRC + '/system.config.js',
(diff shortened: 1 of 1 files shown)


mgechev:
Thanks for the PR!

tools/workflow.config.js
```diff
@@ -34,6 +34,7 @@ var PATH = {
       require.resolve('reflect-metadata/Reflect.js'),
       require.resolve('reflect-metadata/Reflect.js.map'),
       require.resolve('systemjs/dist/system.src.js'),
+      require.resolve('systemjs/dist/system-polyfills.js'),
```
ludohenin: The problem here is that polyfill will be injected in index.html

mgechev: Ups, good point..

ryzy:
Ups, good point indeed ;-)

Any idea how do we want to handle that? Some separate variable in `workflow.config.js` in `PATH {}` there?

ludohenin:
Can be discussed here #120 and here #121

PR #124, opened 2015-10-06
Provide an assets folder

threeaccents:


ludohenin:
not quite sure we will merge that one.

threeaccents:
Okay is it just to keep the seed project as minimal as possible?

mgechev:
@threeaccents yeah, thanks for the PR

PR #127, opened 2015-10-07
Added karma.conf.js to be checked by jshint + fixed errors on that file

tarlepp:


mgechev:
@ludohenin is working on porting our build tasks to TypeScript.

I think we can remove jshint as dependency, `karma.conf.js` is just a single file and I'm not sure it's worth linting.

tarlepp:
Yeah, I noticed that too. But I'm not sure if we should remove jshint task, because some of us still writes plain old javascript.

Maybe some other folder for "generic" tasks?

mgechev:
Since the entire build now uses TypeScript and we will drop jshint, we can close this PR for now.

PR #139, opened 2015-10-11
es6-module-loader, systemjs dependencies bump

ryzy:
Seems like everything works fine after the upgrade.

All references to `es6-module-loader-sans-promises.js` renamed to `es6-module-loader.js`.


mgechev:
Looks good! Thanks!

PR #141, opened 2015-10-11
Allow .spec.js and _spec.js files for writing tests

ryzy:
Many people use Xyz.spec.js for their test files - with this change
we support both, *.spec.js and *_spec.js.

### DIFF
--- a/test-main.js
+++ b/test-main.js
@@ -42,7 +42,7 @@ System.import('angular2/src/core/dom/browser_adapter').then(function(browser_ada
 
 
 function onlySpecFiles(path) {
-  return /_spec\.js$/.test(path);
+  return /[\.|_]spec\.js$/.test(path);
 }
 
 // Normalize paths to module names.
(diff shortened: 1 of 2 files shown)


NathanWalker:
I like having this flexibility, thanks +1

ludohenin:
I'd prefer to have one or the other, having both on a single does not look very clean.
I would not allow it my projects at least :)

ludohenin:
Shall we configure it ?

ryzy:
Hmm, but you guys talking about code guidelines, file naming etc... I'm not sure it's relevant here ;)

For me the most important is that Karma picks the test files, no matter what style someone follows.
To make it configurable would be bit overhead: we have this pattern in 2 places, but it's defined differently (once in a RegExp, once as a glob string). Not sure if it's worth effort.

IMVHO ;)

ludohenin:
@ryzy ya you're right. I think we can merge.

PR #142, opened 2015-10-12
Using @Component and dropping usage of @View

NathanWalker:
Sounds like @View will be deprecated at some point in the future.
Also, just using @Component is more clear, not to mention easier to teach and explain to others.


ludohenin:
Looks good to me. Let's wait for @mgechev thoughts.

mgechev:
Yep, looks good to me as well

PR #144, opened 2015-10-12
Add Jasmine typings

ryzy:
Jasmine is used in the projects, but the typings are missing. This
change adds them.


ryzy:
False alarm, sorry. Jasmine typings are included from angular, so there's no need for that.

My InteliJ didn't index it properly and I thought it's missing :)

PR #152, opened 2015-10-13
Small changes to tasks + added test tasks to use PATH definitions.

tarlepp:


tools/tasks/build.assets.dev.ts
```diff
@@ -1,13 +1,12 @@
 import {join} from 'path';
 import {PATH} from '../workflow.config';
 
-export = function (gulp) {
+export = function task(gulp) {
```
mgechev: Calling the function based on what it does seems more appropriate semantically, i.e. instead of calling the task `task`, you can name it `buildAssets` or something like that. This way if we have a stack trace somewhere we'll get better information about what actually have happened.

tools/tasks/build.index.dev.ts
```diff
@@ -2,9 +2,10 @@ import {join} from 'path';
 import {injectableDevAssetsRef, transformPath, templateLocals} from '../utils';
 import {PATH} from '../workflow.config';
 
-export = function (gulp, plugins) {
+export = function task(gulp, plugins) {
```
mgechev: Same here and below.

tools/tasks/test.dev.ts
```diff
@@ -1,5 +1,8 @@
-export = function (gulp, plugins) {
+import {join} from 'path';
+import {PATH} from '../workflow.config';
+
+export = function task(gulp, plugins) {
   return function () {
-    plugins.watch('./app/**', () => gulp.start('build.test'));
+    plugins.watch(join(PATH.src.all, '**'), () => gulp.start('build.test'));
```
mgechev: Nice catch!

mgechev:
Its great that you fixed the paths! I added some comments we can discuss them if you don't completely agree. If you do, would you push again and squash?

tools/tasks/build.assets.dev.ts
ludohenin: :+1:

tarlepp:
Changed those function names to match with each task file.

PR #161, opened 2015-10-15
Fixed preinstall task, which was failing with Travis.

tarlepp:


ludohenin:
Is that a specific Travis behavior ?

tarlepp:
That's propably because npm install tsd will make local installation to clean linux box, so there isn't tsd installed as global.

mgechev:
Thanks for that! May be we can use path.join for getting the correct path?

tarlepp:
Ah, good catch will check that.

tarlepp:
Hmm, seems like this was not fix...

mgechev:
Yeah, there was a problem with the async behavior of the preinstall task. You can find working build here https://github.com/mgechev/angular2-seed/pull/154

We can close this PR?

ludohenin:
@mgechev I think so

tarlepp:
Thanks for that.

PR #166, opened 2015-10-17
Small fixes.

tarlepp:
1) Removed some not needed ;
2) Some code syntax changes

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -23,64 +23,52 @@ gulp.task('clean.test',  task('clean', 'test'));
 // --------------
 // Postinstall.
 gulp.task('postinstall', done =>
-  runSequence(
-    'clean',
-    'tsd',
-    'npm',
-    done
-  ));
+  runSequence('clean',
+              'tsd',
+              'npm',
+              done));
 
 
 // --------------
 // Build dev.
 gulp.task('build.dev', done =>
-  runSequence(
-    'clean.dist',
-    'tslint',
-    'build.jslib.dev',
-    'build.sass.dev',
-    'build.js.dev',
-    'build.csslib.dev',
-    'build.fonts',
-    'build.index.dev',
-    done
-  ));
+  runSequence('clean.dist',
+              'tslint',
+              'build.jslib.dev',
+              'build.sass.dev',
+              'build.js.dev',
+              'build.csslib.dev',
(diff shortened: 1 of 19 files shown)


mgechev:
Are these changes according to the coding style or just your personal preferences/

tarlepp:
Removing not needed ```;``` is just something that IDE's will complain (if those exists), other things are done according to https://github.com/airbnb/javascript, which is basically standard.

ludohenin:
I find original indentation a lot easier to read. Guidelines are good but redability first imo.

tarlepp:
ok, will revert those identations changes.

mgechev:
Lets close this PR for now.

PS: I just want to mention that redundant use of semicolons is not always a drawback https://github.com/twbs/bootstrap/issues/3057

tarlepp:
ok, imho those should follow same syntax in all files, eg. now there is tasks files with and without these not needed semicolons. Just trying to harmonize those. eg.

https://github.com/mgechev/angular2-seed/blob/master/tools/tasks/build.csslib.dev.ts#L9
https://github.com/mgechev/angular2-seed/blob/master/tools/tasks/build.index.dev.ts#L21

mgechev:
Yes, I agree with you for following a consistent coding style. Semicolons are common for function expressions so this seems like more proper way to proceed https://github.com/mgechev/angular2-seed/blob/master/tools/tasks/build.index.dev.ts#L21.

I can reopen this PR, you can fix all the inconsistencies you find and squash. Does this sounds alright to you?

tarlepp:
ok, will check all those. Also I need to make some default config changes to my IDE (IntelliJ IDEA) for not to complain these "errors", basically IDE is saying:

```
Unnecessary semicolon less... (Ctrl+F1) 
Checks JavaScript source code for unneeded semicolons . The validation works in JavaScript, html 
or jsp files.
```
And that was the reason why I removed those. Semicolons are ok, if you're making eg. ```let somefunction = function () { .... }; but if it's export those are not needed.

tarlepp:
Changes made to that branch. https://github.com/tarlepp/angular2-seed/commit/1ce1dc2e3773abedeb6335dd34df7a3b647f181f

mgechev:
LGTM, would you create a pull request with that branch?

PR #167, opened 2015-10-17
Addded media library task to the build.

arbuckle:
This is a task that pipes the contents of app/media/* to the
target build directory - modeled pretty much exactly like the CSS
handler, but with a path specified that simply traverses the
whole media directory.

Updated the home template and css with a proof-of-concept 
demonstration of this behavior.


mgechev:
LGTM, @ludohenin what are your thoughts?

ludohenin:
I like it too. My only concern is that the more we add, the more to clean-up when starting a project off the seed.

mgechev:
What I don't like with the resource handling right now is that when we need to add another resource type we need to create another task which copies all files of the given type from `app` to `dist/target`.

It'll be better if we had a single task, which copies all the resources and injects only a couple of them, which are explicitly listed.

I see two possible solutions:

- Move all the TS files to separate directory
  + Simplicity
  - The template and all the styles won't longer live together with the component's controller
- Keep the current file structure but filter all TS files and keep them away from the `dist` directory. We will have some ignore list of file extensions which we won't copy to `dist`.
  + We keep the component's controller and all other resources associated with them together
  - More complex

What are your thoughts?

ludohenin:
I like the second solution better. The build might be more complex (to see though) but the component will be easier to reason about. Will look at it.

briantopping:
I had just added #193 when @mgechev noted this PR. 

Long term, I think there ought to be something that is able to look at the different components and find what they individually need copied and to where. Right now, there's no way for a component to have asset requirements that are different than other components. A component should be depending on assets that are in a directory at a higher level directory, or it's not really a component by nature of the lack of containment at the file structure level.

This definitely solves another use case though, one that is needed now.

It doesn't seem like there's any way to solve these issues completely unless the components can hook into the build stages as necessary.

tarlepp:
One solution could be using just one main directory for all of these eg. ```assets``` which could contain subdirectories as devs like.

NathanWalker:
I like @tarlepp suggestion of having one `assets` folder. Simplest and much easier for the build and customization by those who base their projects on the seed.

johnjelinek:
I like `assets`. Do you want a PR for that?

johnjelinek:
#233 is a PR for `assets` that's almost identical to this PR.

NathanWalker:
I agree with @johnjelinek and would opt for his PR over this one.

@arbuckle Thanks for your contribution. For sake of flexibility we may go with @johnjelinek's PR.

/cc @mgechev @ludohenin

mgechev:
@NathanWalker @johnjelinek the things I don't like in the suggestion of having `assets` directory:

- It breaks the component's encapsulation. Currently we add all the resources associated with given component to its own directory (template and CSS). However, if we add assets directory it'll contain resources for any component basically. This looks like coincidental cohesion, I'm not big fan of it.
- What resources we will store in `assets`? Is it only for images and fonts or we can also include CSS files as well?

The disadvantage of this strict encapsulation is that we will have duplication in the styles/images of the individual components (eventually, in a perfect world where components are on low enough level of granularity we will be able to reuse the individual resources because they'll be encapsulated by separate components).

To me, in object-oriented point of view, the best approach is to have the following structure:

```
.
├── components
│   ├── about
│   │   ├── about.css
│   │   ├── about.spec.ts
│   │   ├── about.ts
│   │   ├── fonts
│   │   └── img
│   └── home
│       ├── home.css
│       ├── home.html
│       ├── home.spec.ts
│       ├── home.ts
│       └── img
└── services
    ├── name_list.spec.ts
    └── name_list.ts
```

PR #171, opened 2015-10-19
optimized typings installation

sclausen:
To remove the hen egg problem with typings I migrated the tsd commands from the gulp-task to the beginning of the npm postinstall script, right before gulp executes.


ludohenin:
@sclausen would you please rebase (we have some conflict since I merge alpha42)?

tools/tasks/tsd.ts
```diff
@@ -1,7 +0,0 @@
-export = function tsd(gulp, plugins) {
```
tarlepp: I think that this task is still usefull in some of cases, so I would not remove this.

sclausen:
I might have messed up the commit history a little…

mgechev:
No worries, if you prefer would you open a PR in a fresh branch?

sclausen:
I hope this one is cleaner #173

PR #173, opened 2015-10-20
refactored tsd task

sclausen:
moved tsd task to npm postinstall right before any gulp task and added typings from tsd to `.gitignore`


ludohenin:
LGTM

PR #177, opened 2015-10-20
updated NPM dependencies in json file

jesperronn:
Did not update 'del' to 2.0.2 since that required a change in the 'done'
callbacks for clean.ts

Please note that tests for this commit branch already ran on my fork:

https://github.com/jesperronn/angular2-seed/compare/master...jesperronn:update_dependencies

PR #178, opened 2015-10-21
Revert "updated NPM dependencies in json file"

mgechev:
Reverts mgechev/angular2-seed#177


tarlepp:
What was the reason for this revert?

PR #181, opened 2015-10-22
Fixed broken link.

tarlepp:


README.md
```diff
@@ -149,7 +149,7 @@ npm run karma.start           # 2nd window
 
 # Contributing
 
-Please see the [CONTRIBUTING](https://github.com/mgechev/angular2-seed/blob/master/CONTRIBUTING.md) file for guidelines.
+Please see the [CONTRIBUTING.MD](https://github.com/mgechev/angular2-seed/blob/master/CONTRIBUTING.MD) file for guidelines.
```
mgechev: I think it is better to name it 'CONTRIBUTING.md' in order to be consistent with the readme file. Also lets keep the text associated with the link "Contributing", because it is related to the action rather than the file.

tarlepp:
You're right about that.

PR #185, opened 2015-10-23
Fix #183

### LINKED ISSUE #183: Karma does not load rxjs
Fix path to `@reactivex/rxjs/*`

ludohenin:
Add dependency to `@reactivex/rxjs`


ludohenin:
Looks like test on travis are now well executed :)

PR #191, opened 2015-10-24
Fix issue with IntelliJ

briantopping:
IntelliJ stores transpiles and maps to the current directory. The map files cause the console build to fail unless one clears these files out beforehand. This patch stops that behavior.


briantopping:
Resubmit after cleanup.

PR #192, opened 2015-10-24
Add images copy

briantopping:
Add copy images functionality. Maybe this should be merged with fonts and other generic copy tasks.


briantopping:
Meh, bad rebase

Issue #158, opened 2015-10-14
Add clean node_modules folder onpreinstall ?


ludohenin:
I've also had some troubles installing the deps on a non fresh clone.
I think it'd good to clean node_module folder in preinstall script.

tarlepp:
I think this would solve some issues about installation.

tarlepp:
We could use same approach as in ng2 repository: https://github.com/angular/angular/blob/master/package.json#L26

mgechev:
@tarlepp makes sense

ludohenin:
@tarlepp :+1: - Shall I assign it to you ?

tarlepp:
Seems like npm shrinkwrap is the way to handle this. This will cause some "extra" work when updating packages but not much. More info https://github.com/angular/angular/blob/master/npm-shrinkwrap.readme.md .

Bigger problem is npm itself, seems like 3.x version has some major problems with shrinkwrap, more info https://github.com/npm/npm/issues/9907

mgechev:
Yeah, I have some painful memories with shrinkwrap...Lets stay away from that until it gets the only option :-)

tarlepp:
You have some other solution / thoughs for this?

mgechev:
We can clean node_modules by using fs.unlinkSync for example.

tarlepp:
sure but that won't fix the problem, also that would not prevent package versions changes. If that is the only option, then I would just add some "installation problems" section to readme.md and describe this there.

mgechev:
Why would we need to lock the versions of the dependencies of the dependencies of the seed?

@ludohenin your problem occurs because of some mess around overriding of the dependencies caused by preinstall with existing `node_modules`, right?

tarlepp:
imho the reason are fairly clear explained on ng2 repo https://github.com/angular/angular/blob/master/npm-shrinkwrap.readme.md

If we don't want to _or_ can't use this aproach we should add eg. "Possible failures" to README.MD and provide these solutions there.

ludohenin:
> If we don't want to or can't use this aproach we should add eg. "Possible failures" to README.MD and provide these solutions there.

I'd prefer the possible failure approach (a lot simpler :).
isn't it enough what we already have in the readme ?
```bash
npm install   # clean npm cache & delete node_modules folder if you get an error
```

jesperronn:
[...]

Issue #176, opened 2015-10-20
Tsd reinstall often fails with 'Github rate exceeded'


jesperronn:
This build, for instance: https://travis-ci.org/mgechev/angular2-seed/jobs/86486664#L156

Fails with

```
tsd reinstall --clean && tsd link && tsd rebundle && gulp postinstall -> running reinstall
    [ERR!] GitHub API: GitHub rate limit reached.
    To increase the limit use GitHub authentication.
    See: https://github.com/DefinitelyTyped/tsd#tsdrc
```

Documentation itself suggests to add a `token` into `.tsdrc`. Which is kind-of weird to have on Travis/CircleCI since it's a private github key.

Imagine if npm dependencies were pulled from Github. A CI provider would easily spend the 60 api requests per hour. Forgive my lack of knowledge, but why is tsd pulling this from Github and precicely what is it pulling from Github?

Is there a way that stuff inside the `tsd` command can be cached? Or pulled from somewhere else?

tarlepp:
This can be solved simply by create of auth token for tsd. This is described in https://github.com/DefinitelyTyped/tsd#tsdrc

\\CC @mgechev You should create this token as described and then create ENV variable to Travis configuration.

jesperronn:
I understand. But I have the impression that these tokens are personal and
therefore should not be committed into repository. Is this correct?

For instance if I create a token with my github account, then this token
can be used to make other API requests on my behalf. I should never submit
that into this repository (even thought I was pretty close doing that).

Is this correctly understood?

Also, if you have links explaining extactly what `tsd` is doing and why it
needs to access github at this point??? I don't understand, and assume it
must slow the build down.


On Wed, Oct 21, 2015 at 5:04 PM, Tarmo Leppänen <notifications@github.com>
wrote:

> This can be solved simply by create of auth token for tsd. This is
> described in https://github.com/DefinitelyTyped/tsd#tsdrc
>
> \CC @mgechev <https://github.com/mgechev> You should create this token as
> described and then create ENV variable to Travis configuration.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mgechev/angular2-seed/issues/176#issuecomment-149925312>
> .
>



-- 

Jesper Rønn-Jensen
Nine A/S
Mobile: +45 2989 1822
Blog http://justaddwater.dk/
jesperrr@gmail.com (Private e-mail and Google Talk IM)

tarlepp:
[...]

PR #197, opened 2015-10-26
If npm install fails, use "npm run reinstall" to ensure we start at clean slate

jesperronn:
This addresses issue #158, but only when you are in trouble.

This approach has a tradeoff: Firstly it is not done automatically, which requires the use to think after being in trouble

On the other hand, this approach saves time since a full clearing of caches and installation of all modules takes time.

Furthermore, adding the command as "npm run reinstall" makes it explicit what should and could be done

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ It is something similar to the AngularJS Quick Start but does the entire build w
 ```bash
 git clone https://github.com/mgechev/angular2-seed.git
 cd angular2-seed
-npm install   # clean npm cache & delete node_modules folder if you get an error
+npm install   # or `npm run reinstall` if you get an error
 npm start     # start with --env dev
 ```
 _Does not rely on any global dependencies._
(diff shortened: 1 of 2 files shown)


jesperronn:
Travis error is unrelated.

See #176 `tsd reinstall` fails with 'Github rate exceeded'

mgechev:
@jesperronn LGTM. Would you revert back the README? I'm generating the contributors' list with https://github.com/mgechev/github-contributors-list

jesperronn:
@mgechev sorry, my bad! I had a markdown table formatter installed in my editor. New update on its way

jesperronn:
@mgechev if you don't mind, I will do a 'push --force' to replace history in this branch. 

(pretty history after merge)

mgechev:
Yes, sounds good

jesperronn:
done, @mgechev, merge?

mgechev:
Yep, looks good! I'll update the contributors list tomorrow.

Issue #137, opened 2015-10-10
Add PhantomJS2 to Karma's browsers


ryzy:

ryzy:
I'm trying to come up with working code for that, but have a problem with PhantomJS 2.0 throwing some weird errors.

Here are the changes: https://github.com/ryzy/angular2-seed/commit/4d14f5f07825274043003439a3428a8760792f74

When I run it, I have following result:
```
[20:26:28] Starting 'karma.start'...
INFO [karma]: Karma v0.12.37 server started at http://localhost:9876/
INFO [launcher]: Starting browser PhantomJS2
INFO [launcher]: Starting browser Chrome
INFO [PhantomJS 2.0.0 (Mac OS X 0.0.0)]: Connected on socket fN-19W3eEMiVRol7N_p8 with id 74726523
PhantomJS 2.0.0 (Mac OS X 0.0.0) ERROR: 'TypeError: undefined is not a constructor (evaluating 'new exports.Map()')
	Evaluating http://localhost:9876/base/node_modules/angular2/src/core/facade/collection.js
	Evaluating http://localhost:9876/base/node_modules/angular2/src/core/dom/browser_adapter.js
	Error loading http://localhost:9876/base/node_modules/angular2/src/core/dom/browser_adapter.js'

PhantomJS 2.0.0 (Mac OS X 0.0.0): Executed 0 of 0 ERROR (0.001 secs / 0 secs)
PhantomJS 2.0.0 (Mac OS X 0.0.0): Executed 0 of 0 ERROR (0.001 secs / 0 secs)
Chrome 45.0.2454 (Mac OS X 10.11.1): Executed 3 of 3 SUCCESS (0.083 secs / 0.079 secs)
[20:26:31] 'karma.start' errored after 2.52 s
```

Any ideas what might be wrong? Still quite new with ES6/TS/Angular2.

tarlepp:
Yeah, I tried that too but I got some referenceErrors about missing 'require'. Going to try this one again...

ryzy:
@tarlepp there were some missing references and even without PhantomJS there were warning. I fixed it in that change: https://github.com/ryzy/angular2-seed/commit/c21c7cdb9804219f8829c376574a79859a4ce391

Will do PR with these two commits once I managed to make PhantomJS work...

tarlepp:
With just installing ```karma-phantomjs2-launcher``` and changing the ```karma.conf.js``` I got following.

```bash
[...]

PR #198, opened 2015-10-27
WIP Removing traceur, which is no longer needed for Angular2

jesperronn:
WIP, this fails, so it must be fixed before it should be merged

See https://github.com/angular/angular/issues/3737#issuecomment-132877615

> "now that Traceur is gone we should be able to fix this"


jesperronn:
According to https://github.com/mgechev/angular2-seed/issues/137#issuecomment-147175881

> Seems like PhantomJS2 cannot work without Traceur - after restoring its
> dependency (which was removed in  _Update to angular@2.0.0.alpha.38_
> commit: mgechev@b94031c)

Which means that we cannot remove Traceur anyway.

Issue #184, opened 2015-10-23
Keep tsd definitions inside the repo ?


ludohenin:
I see a couple of advantages to do so:

* No more error with Travis (or less at least)
* Already referenced `./tools/typings/tsd/**/*.d.ts` in `tsconfig.json` files
  * right now if we just clone, install, start ... tsd definition are missing in tsconfig

tarlepp:
Travis is failing because of GitHub rate limit, which can be easily solved as described in #176 

If the point is keep this repo as barebone as possible, then I wouldn't include these typings. And imho it's good that repo contains only things that *we* have coded, not 3rd party libraries.

But generally I think there is no "wrong" or "right" answer for this.

ludohenin:
My main concern here is 

> right now if we just clone, install, start ... tsd definition are missing in tsconfig

sclausen:
@ludohenin The problem that the typings are missing in `tsconfig.js` could be solved with adding the following:
``` json
{
    "filesGlob": [
        "!./tools/typings/tsd/**/*.d.ts"
    ]
}
```
I don't know how other IDE's treat the `tsconfig.json`, but atom-typescript doesn't mark e.g. any express imports as errors (`tools/utils/server.ts`) If I exclude tsd-typings in filesGlob. Would this be a valid solution for one of your arguments?

IMHO including the tsd typings in git is really dirty project design. It adds unnecessary files to the repository, because their versions are already defined in `tsd.json`. One wouldn't include node_modules, or bower_components either.
I think doing this just because of the git rate limit and travis is not worth the cons.

Issue #169, opened 2015-10-19
Refactor: add tsd-files defined in tsd.json to gitignore


sclausen:
Tracking them is IMHO unnecessary and bloats the git repo.

ludohenin:
@sclausen this is because installing typings on npm preinstall is not stable considering we don't want to rely upon tsd being globaly installed.
Btw If you have a working solution (working on Travis as well) that'd be awesome :)

sclausen:
Installing typescript definitions has to be done before compiling any typescript and you can't use the local tsd before installing it, so why not moving the tsd commands in the gulp-task to to the postinstall npm task?
``` json
    "postinstall": "tsd reinstall --clean; tsd link; tsd rebundle; gulp postinstall",
```
``` typescript
gulp.task('postinstall', done =>
  runSequence('clean',
              'npm',
              done));
```
now you would execute tsd commands after installing the node modules, but before executing any gulp scripts and you could add `tools/typings/tsd` to the `.gitignore`.

ludohenin:
Is this (`"postinstall": "tsd reinstall --clean; tsd link; tsd rebundle; gulp postinstall",`) cross platform ?

sclausen:
:frowning: I didn't thought there could be any compatibility problems. Unfortunately, I don't have multiple os on hand, so how could I test this?

sclausen:
I tested it on travis-ci and it succeeds.

ludohenin:
It works on OSX for me.
Anyone who can test on Windows ?

svrcekmichal:
@ludohenin it works :) i prefer `&&` syntax, but it's running ok

ludohenin:
Good to read!I also prefer `&&` but wasn't sure if it'd run across platforms.
@sclausen would you send a PR (single commit & with latest master) ?

sclausen:
@ludohenin Here you go #171

sclausen:
closed with #173

PR #200, opened 2015-10-27
Fix build errors by explicitly adding tsd files in the project

### LINKED ISSUE #176: Tsd reinstall often fails with 'Github rate exceeded'

jesperronn:
Fixes #176

The pull request #173 introduced a lot of builds failing in errors because Tsd task would attempt to retrieve typings files from the https://github.com/borisyankov/DefinitelyTyped repository.

The problem is that CI builds have to use the network to retrieve the typings files, and this fails often (see discussion in #176).

https://github.com/DefinitelyTyped/tsd/#notes notes that

> It is recommended you check-in the definitions you install into your VCS.

The commit here adds all the typings files to this repository. Which is recommended to do. This is a tradeoff, since the "tsd" task must run at any point in time where we update dependencies or change type definitions.

On the other hand, build tasks can run without using the network.

The final point I am trying to make is that I would like to avoid the flakyness, and introduce as much stability in the building of the seed project as possible.

For discussion of the pros and cons, see

https://github.com/mgechev/angular2-seed/pull/173
https://github.com/mgechev/angular2-seed/issues/176
https://github.com/mgechev/angular2-seed/issues/169
https://github.com/mgechev/angular2-seed/issues/171
https://github.com/mgechev/angular2-seed/issues/184


package.json
```diff
@@ -12,12 +12,11 @@
     "karma": "karma",
     "karma.start": "karma start",
     "lint": "gulp tslint",
-    "postinstall": "tsd reinstall --clean && tsd link && tsd rebundle && gulp postinstall",
     "start": "gulp serve --env dev",
     "serve.dev": "gulp serve --env dev",
     "tasks.list": "gulp --tasks-simple",
     "test": "gulp test",
-    "tsd": "tsd"
+    "tsd": "gulp tsd"
```
ludohenin: I disagree with this as `npm run tsd -- <args>` is just a proxy to `tsd` bin for those who don't want to have it installed globally.

ludohenin:
:+1: (except my comment)

package.json
jesperronn: @ludohenin OK, I understand. Beware, though that `gulp tsd` is a proxy for the previous `tsd reinstall --clean && tsd link && tsd rebundle`. 

Should I create another task (for easy discovery)? If so, what name? 

Suggestion: `npm run tsd:reinstall` or `npm run tsd:update`

jesperronn:
@ludohenin rewrite done. Wait for tests be finished.

ludohenin:
I think we should close that one. I was in favor of it but I've to admit it's the cleanest way to do things.
@mgechev any thoughts ?

mgechev:
Agree with you. I'm not a big fan of tsd either but I don't see better options.

jesperronn:
@ludohenin @mgechev I accept your opinions for this, although I still see the problems for the build failures all over. 

You are already aware that Travis will not resolve the env parameters for pull requests, due to security reasons. This means that all pull-requests will still fail as long as tsd files are not in the repos. 

This in turn makes it harder for you to consider external contributions. 

Again, it's your project, your call.

tarlepp:
Actually Travis should follow the ENV variable that is set it from Travis Setting GUI. I have set that from there and no problems at all after that. And that ENV variable is used when PR is made and Travis starts to run those tests.

Issue #195, opened 2015-10-26
npm postinstall fails in node 0.12


jimthedev:
I get this error when running npm install in node 0.12

```

➜  angular2-seed  master ✓  npm run postinstall

> angular2-seed@0.0.0 postinstall /Users/jcummins/projects/angular2-seed
> tsd reinstall --clean && tsd link && tsd rebundle && gulp postinstall


-> running reinstall

>> written zero files

   >> definition-header (node)
   >> reflect-metadata (node)

>> nothing rebundled
[12:02:17] Requiring external module ts-node/register

/Users/jcummins/projects/angular2-seed/tools/utils/tasks-tools.ts:32
    readdirSync(path).forEach(function(file) {
                      ^
TypeError: undefined is not a function
    at /Users/jcummins/projects/angular2-seed/tools/utils/tasks-tools.ts:38:47
    at Array.forEach (native)
    at walk (/Users/jcummins/projects/angular2-seed/tools/utils/tasks-tools.ts:32:23)
    at scanDir (/Users/jcummins/projects/angular2-seed/tools/utils/tasks-tools.ts:29:3)
    at Object.autoRegisterTasks (/Users/jcummins/projects/angular2-seed/tools/utils/tasks-tools.ts:10:3)
    at Object.<anonymous> (/Users/jcummins/projects/angular2-seed/gulpfile.ts:13:1)
    at Module._compile (module.js:460:26)
    at Object.loader (/Users/jcummins/projects/angular2-seed/node_modules/ts-node/dist/ts-node.js:94:18)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)

npm ERR! Darwin 14.5.0
npm ERR! argv "node" "/Users/jcummins/bin/npm/bin/npm" "run" "postinstall"
npm ERR! node v0.12.0
npm ERR! npm  v3.3.8
npm ERR! code ELIFECYCLE
npm ERR! angular2-seed@0.0.0 postinstall: `tsd reinstall --clean && tsd link && tsd rebundle && gulp postinstall`
npm ERR! Exit status 1

```

tarlepp:
Hmm, I think that we need to also update ```travis.yml``` to run all checks with 0.12 node.

ludohenin:
@tarlepp :+1:

mgechev:
We support npm 3.x.x & node 4.x.x. Here's the version assertion @tarlepp implemented https://github.com/mgechev/angular2-seed/pull/216/files

PR #196, opened 2015-10-26
fix(build): support node 0.12

jimthedev:
This adds support for node 0.12 in tools so that the postinstall completes

Closes #195

PR #203, opened 2015-10-27
Update .travis.yml

tarlepp:
Added ```node 0.12``` to be also included in tests. #195

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 
 node_js:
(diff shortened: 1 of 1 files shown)


tarlepp:
Tests are currently failing with ```0.12``` version because the fix from #196 isn't yet merged. This should prevent similar failures in future.

tarlepp:
This one can be closed because of #207 which solves this one too.

PR #222, opened 2015-11-02
Major Travis CI update

tarlepp:
Added support for both Linux and OSX environments
Added support for following node versions;
 * 4.0.x
 * 4.1.x
 * 4.2.x
 * 5.0.x

Support for node 0.12.x is commented atm, waiting for #196 to be merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,13 @@ addons:
 
 env:
   matrix:
-  #- TRAVIS_NODE_VERSION="0.12" // Enable this line when #196 is merged
-  - TRAVIS_NODE_VERSION="4.0"
-  - TRAVIS_NODE_VERSION="4.1"
-  - TRAVIS_NODE_VERSION="4.2"
-  - TRAVIS_NODE_VERSION="5.0"
+    #- TRAVIS_NODE_VERSION="0.12" // Enable this line when #196 is merged
+    - TRAVIS_NODE_VERSION="4.0"
+    - TRAVIS_NODE_VERSION="4.1"
+    - TRAVIS_NODE_VERSION="4.2"
+    - TRAVIS_NODE_VERSION="5.0"
+  global:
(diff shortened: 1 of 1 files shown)


tarlepp:
Hmm GitHub rate limit is still present for some reason? Could @mgechev or @ludohenin confirm that Travis has that ENV variable set?

Below is link to Travis build which I used to test this PR:
https://travis-ci.org/tarlepp/angular2-seed/builds/88853718

mgechev:
@tarlepp updated [here](https://github.com/mgechev/angular2-seed/commit/76e0edc523697a25e392d15653faa9943c357222#diff-354f30a63fb0907d4ad57269548329e3R22).

There's a merge conflict. Would you resolve it and push again?

tarlepp:
Hmm, you don't need to add that ENV variable to travis.yml, there is a simple setup in travis itself for that so that those env variables are not to shown to anyone.

image

tarlepp:
And seems like that env -> global -> secure thing is not working, hitting rate limit once again.

jesperronn:
@tarlepp @mgechev Adding the Travis Env variable will not work for pull requests. this is due to security reasons: (A malicious contributor could request a PR printing all secret API keys in the console).

Therefore, the Github Rate issue prevails. And it will raise the bar for all contributors. 

I am currently seeing this error all over the place in random pull requests. And it's really a shame, since it prevents core team from seeing real problems in the code.

@mgechev @ludohenin I know you are aware of this problem, and aware that the current way to solve it is not requesting TSD files during build.

(as a side note, perhaps if Microsoft would serve the TSD files just like NPM does today, would help in the future. But for now, Github will not accept being used as a fileserver for TSD files)

We have discussed this previously. One fix: Add tsd files to repository #200 

Issues and pull requests affected are numerous. Here are some others where we discuss:

* #176
* #184

mgechev:
@tarlepp can you verify if the github token works now?

tarlepp:
You could restart those test from https://travis-ci.org/mgechev/angular2-seed/builds/88882783 upper right corner you should see "re-run" icon. I can't do that because I'm not owner of that Travis instance.

mgechev:
@tarlepp this is not going to pass anyway since in your (updated) version of `travis.yml` doesn't contain the token.

tarlepp:
going to give up with this :(

mgechev:
It looks great to me! Just add these lines to your file and I think the builds should pass https://github.com/mgechev/angular2-seed/blob/master/.travis.yml#L20-L24

tarlepp:
This could be solved just by adding that same token via Travis CI in your settings page, as I have said many times...

Going to make this PR again at some point.

mgechev:
The travis token is in the settings page already.

Issue #219, opened 2015-11-01
How do I update tsconfig.json


jesperronn:
More of a question. How exactly do I keep my `tsconfig.json` file updated?

I tried running `npm run postinstall`, but I am not sure it catches all changes. 

(postinstall job runs `tsd reinstall --clean && tsd link && tsd rebundle` and more)

PS> I tried to add a .ts file somewhere, and saw that no line was added to the `tsconfig.json` file. But when I tried to update `package.json`, then the file updated.

What is supposed to update the tsdconfig.json, and how does that happen.

mgechev:
Some IDEs and text editors does this automatically.
I have the same issue since I'm using vim. I'd suggest using paths with wildcards in the files array.

The problem comes when a new files array is being generated by the IDE/text editor of another contributor. In such case what we have here in the repo will be overwritten. How about, updating the tsconfig.json to something like:

```json
{
    "version": "1.6.0",
    "compilerOptions": {
        "target": "es5",
        "module": "commonjs",
        "declaration": false,
        "noImplicitAny": false,
        "removeComments": true,
        "noLib": false,
        "emitDecoratorMetadata": true,
        "experimentalDecorators": true,
        "sourceMap": true
    },
    "files": [
        "app/**/*.ts",
        "gulpfile.ts",
        "tools/**/*.ts"
    ],
    "filesGlob": [
        "./app/**/*.ts",
        "./gulpfile.ts",
        "./tools/**/*.ts"
    ],
    "compileOnSave": false,
    "exclude": []
}
```

After we update it to the content above we can add it to `.gitignore` and prevent it from further changes.

ludohenin:
Looks good.
What about this:
```
{
    "version": "1.6.0",
    "compilerOptions": {
        "target": "es5",
        "module": "commonjs",
        "declaration": false,
        "noImplicitAny": false,
        "removeComments": true,
        "noLib": false,
        "emitDecoratorMetadata": true,
        "experimentalDecorators": true,
        "sourceMap": true
    },
    "files": [
        "**/*.ts"
    ],
    "filesGlob": [
        "./**/*.ts"
    ],
    "compileOnSave": false,
    "exclude": ["node_modules"]
}
```

mgechev:
Yes, I like it

jesperronn:
@mgechev @ludohenin Even though `files` and `filesGlob` are simpler in your suggestions, how can we get closer to this example:
[...]

PR #224, opened 2015-11-03
Tsconfig simplification: No need to specify all files.

jesperronn:
* [x] Specify outdir so that IDE builds will put generated js there.
      This allows Atom editor's Typescript plugin to work without dropping
      tsconfig.json files and generated .js files in the source directory

* [x] simplify tsconfig.json so that it is NOT updated by an eager
      IDE if files are added/removed

This fixes issue #219 and is following how Angular2 does it.

See also angular2 commit https://github.com/angular/angular/commit/5b104936ae55dc1b5b5ca687a912e703dd50dfa4 by @alexeagle'


ludohenin:
I don't think we need the outDir to be set. We have a task for compiling ts.

jesperronn:
@ludohenin OK will remove, then

ludohenin:
I'd also keep `compileOnSave` to false.

jesperronn:
@ludohenin I removed it because `compileOnSave` is not in the spec

PR #245, opened 2015-11-13
Fix #244

### LINKED ISSUE #244: npm run reinstall broken on windows
```
'rm' is not recognized as an internal or external command, operable program or batch file.
```

I think we have to either use rimraf or a task for that.

ludohenin:
Add rimraf for cross platform compatible unix `rm -rf` command.


ludohenin:
looks like 1min isn't enough for appveyor :-/

PR #247, opened 2015-11-14
fixing issue that prevents more than one test using Router

larsthorup:
Moving beforeEachProviders() from global scope, where it only makes sense to have one, to inside the describe. Then other tests can have their own beforeEachProviders() inside their describes and Jasmine will take care of cleaning up the installed providers after the describe has been run.


ludohenin:
Looks good, thanks!

PR #248, opened 2015-11-15
Integration with materializecss and added contributors page

ultrasonicsoft:
Hi Minko,
I have done below changes:
1. integrated website with materializecss
2. updated about page
3. added contributors page

### DIFF
--- a/app/components/about/about.html
+++ b/app/components/about/about.html
@@ -1,13 +1,4 @@
 <div class="row">
-</div>
-
-<div class="row">
-  <div class="col s12 m6">
-    <a class="waves-effect waves-light btn-large" href="/"><i class="material-icons left">view_headline</i>Home</a>
-  </div>
-</div>
-
-<div class="row">
   <div class="col s12 m6">
     <div class="card blue-grey darken-1">
       <div class="card-content white-text">
--- a/app/components/about/about.html
+++ b/app/components/about/about.html
@@ -1,14 +1,38 @@
-<p>
-  For reward, here is a list of awesome computer scientists!
-</p>
+<div class="row"></div>
+<div class="row">
+  <a class="waves-effect waves-light btn" href="/"><i class="material-icons left">view_headline</i>Home</a>
+</div>
 
-<p>
-  You want more? Add them yourself!
-</p>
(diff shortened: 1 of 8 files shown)


mgechev:
Using material design sounds good to me.

However, including the styles & scripts from CDN seems unnecessary. Bootstrap can be replaced with material and added to npm.

I also like the idea of having a component, with all the contributors but it seems like a better option to generate it dynamically ([just like we do for the readme](https://github.com/mgechev/github-contributors-list)). I'm not sure there's both a good solution to this problem and keeping the project decoupled from the GitHub API/github-contributors-list module.

ultrasonicsoft:
I have planned to generate them dynamically and pushing material design in assets folders for later commits.
I thought first will do check-in of working prototype and then will refactor it again.

One more thing, when i added material folder in my app folder, i was unable to update gulpfile to copy this folder into dist folder.
Can you tell me where should I do changes in gulpfile so that if I have folder say "materializecss" in app folder, after each build it should be copied into dist folder?

Also, are you going to merge this PR or you want me to commit these with dynamic contributors list?

ultrasonicsoft:
In my opinion Material has rich UX than bootstrap, thus went for it. For example, cards, input with moving label etc.
Have you ran the website with my commits? I can provide screen shots if you want?

ultrasonicsoft:
image

image

image

mgechev:
How do you plan to generate the contributors list dynamically for the application?

Do you think the dynamic generator is supposed to live in the build process or you're going to update the list manually when new contributor is added?

There are also other problems since the builds on Travis and AppVeyor are failing so I won't be able to merge the commit until the build passes.

ultrasonicsoft:
github has apis  to fetching information. Below is one for getting contributors dynamically. 
https://developer.github.com/v3/repos/#list-contributors

So can create a contributors service which will hit above url, parse result and show on page, I guess.

Anyhow, it's not something great changes which must be integrated. I had free time on Sunday and thought I can contribute to this project as it helped me lot.

I am fine with any strategy or suggestions you have.

mgechev:
I don't think it is a good idea to couple the seed with the GitHub's API. Lets close it for now.

PR #252, opened 2015-11-18
Prevents dist, doc, lib, .. dirs from being excluded from node_modules.

bertvh:
In case you want to commit node_modules, the existing excludes (dist, lib, etc...) are problematic because they also exclude these dirs inside node modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,10 +30,11 @@ node_modules
 
 #IntelliJ configuration files
 .idea
+.iml
 
-dist
-dev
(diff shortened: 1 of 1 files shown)


mgechev:
`node_modules` is in `.gitignore` already so if you need to push the `node_modules` directory you'd already need to use custom `.gitignore`. I'm not sure if this change is necessary, lets close the issue for now and reconsider it if there are another developers who need this.

PR #278, opened 2015-12-05
Update to angular Alpha 48 (broken)

jesperronn:
This breaks since tests cannot find "rxjs/operators/toPromise" which
changed to "rxjs/operator/toPromise" (singular)

Also, only 2 tests run right now, which also must be fixed

I believe that the first fix requires a new Angular release.
See angular/angular#5641 for discussion


mgechev:
@jesperronn we'll first need to merge alpha47 and tag it. Right after that we can merge alpha48.

sclausen:
:+1:

PR #280, opened 2015-12-07
Remove assets task (load from src)

ludohenin:


ludohenin:
We need to work with source (not dest) in the prod build.

PR #287, opened 2015-12-10
Add fancier error message when npm or node are not required version

mgechev:
Two changes:
- Made the code in `walk` compatible with older versions of node (in order to allow the user to see the error message).
- Add chalk for cleaner error message that the current node or npm version is not supported.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   "devDependencies": {
     "@reactivex/rxjs": "^5.0.0-alpha.6",
     "async": "^1.4.2",
+    "chalk": "^1.1.1",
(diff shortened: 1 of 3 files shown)


rahul71:
But how do I resolve this issue? Isn't the correct version of node supposed to be a part of package.json?

Bigous:
@rahul71 We get into a problem, because package.json engine version is to "run" version, but the seed, after the build will run inside a browser, not in node, so, technically, we should not set node version at package.json.
Now, how to resolve, upgrade node, or use nvm to have multiple node instalations and use 4.0.0 or superior to build this seed.

PR #294, opened 2015-12-11
Performance improvements in the build

mgechev:
The current build in master is very slow because of:
- Introduced new set of tasks for the production build
- `ts-node` 5.4 seems a bit slower

Loading tasks on demand. Improve the initial build from 14 seconds to 8 seconds. Also provides better feedback to the user.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -26,7 +26,7 @@ module.exports = function(config) {
       'node_modules/reflect-metadata/Reflect.js',
 
       { pattern: 'node_modules/angular2/**/*.js', included: false, watched: false },
-      { pattern: 'node_modules/@reactivex/rxjs/dist/**/*.js', included: false, watched: false },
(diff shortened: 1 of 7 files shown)


tarlepp:
Nice speed improvement. Also I think that it would be very nice if actual watch task would only trigger necessary changes. Eg. if user modifies some .sass file there is no need to build whole project.

ludohenin:
@tarlepp Incremental build will surely come soon :)

ludohenin:
`npm start` starts the server but not the `watch` task

mgechev:
Yep, thanks for the catch! The arrow function was creating troubles. It should be fine now (although one second slower :sob:)

ludohenin:
Arrggfff ... there's an issue with `npm test` now. I can reproduce what travis' reporting.

mgechev:
I'll fix that PR once I finish playing with the hot loader :-).

mgechev:
@ludohenin are you still able to reproduce the `npm test` issue?

PR #311, opened 2015-12-17
Fix #310

### LINKED ISSUE #310: Injecting js does not work in Windows
Node: v4.2.3
npm: 2.14.7
OS: Windows 7

Injection of dependencies into index.html does not work in Windows. 
I noticed that when running "gulp build.prod -- env prod". No JavaScript was injected at all.
For development build, it works, but this seems to be "by accident".

For me, it worked after the following changes to build.index.ts:

import {join, **sep**} from 'path';

...
function mapPath(dep) {
    console.log('inject dep: ' + dep);
    let prodPath = '${dep.dest}/${dep.src.split(**sep**).pop()}';
....

mgechev:


ludohenin:
Not critical but I've a tslint error
```
...\angular2-seed\tools\utils\tasks_tools.ts
line 23  col 1  trailing whitespace
× 1 error
```

ludohenin:
Otherwise, it works OK on my Windows machine.

PR #312, opened 2015-12-17
Load only necessary task files

ludohenin:
* Improve build start by just loading the required tasks.
* Single tasks like `check.versions` now needs to be explicitly declared in gulpfile (as not all the tasks file are loaded on start).


lihaibh:
Every "group" of tasks must use this 'runsequence' method you created? I say there is a better solution you can load the tasks which returns a promise, each task loaded only when it is called, so in "task" method we will return function(){ return require(...)(gulp...) }

PR #313, opened 2015-12-18
Lazy load gulp tasks

lihaibh:
According to pull request 312, ludohenin created a "new" runSequence method for group tasks in order to load the proper files before we run the task.
I think its unnecessary, and it is better to load everything, but just using a cb.
When the task runs every subtask will load itself.. gulp invoking the cb => we require the file..
it is also not possible to run specific tasks right now like: gulp build.index, and "clean" tasks also pre-loaded, this change will fix that.

### DIFF
--- a/tools/utils/tasks_tools.ts
+++ b/tools/utils/tasks_tools.ts
@@ -2,35 +2,21 @@ import * as gulp from 'gulp';
 import * as util from 'gulp-util';
 import * as chalk from 'chalk';
 import * as gulpLoadPlugins from 'gulp-load-plugins';
-import * as _runSequence from 'run-sequence';
 import {readdirSync, existsSync, lstatSync} from 'fs';
 import {join} from 'path';
 import {TOOLS_DIR} from '../config';
 
 const TASKS_PATH = join(TOOLS_DIR, 'tasks');
 
-// NOTE: Remove if no issues with runSequence function below.
-// export function loadTasks(): void {
-//   scanDir(TASKS_PATH, (taskname) => registerTask(taskname));
-// }
-
-export function task(taskname: string, option?: string) {
-  util.log('Loading task', chalk.yellow(taskname, option || ''));
-  return require(join('..', 'tasks', taskname))(gulp, gulpLoadPlugins(), option);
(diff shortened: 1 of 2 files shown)


mgechev:
@ludohenin would you take a look?

@lihaibh if we merge would you squash into a single commit?

lihaibh:
@mgechev there was no changes on 'master' branch so why wouldnt u be able to merge? i can still squash all the 4 commits into 1 though..

mgechev:
Sure! Thank you for your PR and your effort!

We're able to merge, also the checks are passing, however, the code should go through a review process. Since @ludohenin is on charge of this feature/improvement he is supposed to take a look at the code.

Don't offend that we're going to review your changes, it is definitely not anything personal. Although we have write access to master we're still enforcing the review process even for our changes in order to ensure quality. :-)

tools/utils/tasks_tools.ts
```diff
@@ -9,14 +9,15 @@ import {TOOLS_DIR} from '../config';
...
 export function task(taskname: string, option?: string) {
-  util.log('Loading task', chalk.yellow(taskname, option || ''));
-  return require(join('..', 'tasks', taskname))(gulp, gulpLoadPlugins(), option);
+  return () => {
+    util.log('Loading task', chalk.yellow(taskname, option || ''));
+    return require(join('..', 'tasks', taskname))(gulp, gulpLoadPlugins(), option)();
+  };
 }
 
 export function runSequence(...sequence: any[]) {
```
ludohenin: I think we can remove this function then.

lihaibh: Yes i think its useless for now also if u think about that, if u call the function with (x,y,z) if z is invalid task name it will still work.. it will just ignore it and run what it actually finds. So i remove your method.

ludohenin:
You can also remove this line https://github.com/mgechev/angular2-seed/blob/master/gulpfile.ts#L11

ludohenin:
@lihaibh would you squash into one commit ?

mgechev:
@ludohenin mergining?

Issue #318, opened 2015-12-21
Broken project compilation on change


tandu:
With this revision 845772264ef64ad5fcc1a33293d714962ced9a63 when I save my changes project is not rebuilt automatically and I need to stop npm process and restart it again in my terminal to rebuild it.

mgechev:
Thanks for reporting the issue! The issue is fixed by https://github.com/mgechev/angular2-seed/commit/8fd110f4b825839af7e5ced634d69fe7f8973bf9

PR #319, opened 2015-12-22
Revert "Lazy load gulp tasks"

mgechev:
Reverts mgechev/angular2-seed#313

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -1,10 +1,5 @@
 import * as gulp from 'gulp';
-import * as runSequence from 'run-sequence';
-import {loadTasks, task} from './tools/utils';
-
-// --------------
-// Configuration.
-loadTasks();
+import {runSequence, task} from './tools/utils';
 
 // --------------
 // Clean (override).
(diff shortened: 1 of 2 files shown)


ludohenin:
:)

lihaibh:
?

mgechev:
@lihaibh https://github.com/mgechev/angular2-seed/issues/318

stefanocke:
@lihaibh , @mgechev  I think the reason for #318 is that the build tasks were made asynchronous "twice" by the PR: They already returned pipes before and you added the "done" parameter, but you never called done(). I am by far not a gulp expert, but I tried the following and it seems to work so far:

```
export function task(taskname: string, option?: string) {
  return () => {
    util.log('Loading task', chalk.yellow(taskname, option));
    return require(join('..', 'tasks', taskname))(gulp, gulpLoadPlugins(), option)();
  };
}
```
So, compared to the PR, I just removed the "done" parameter.

stefanocke:
P.S.: Sorry, spoken a bit to early: My approach does not work for the tasks, that don not return pipes and expect the done parameter, for example build.bundles. Nevertheless: Passing the "done" parameter into tasks, that do not expect it, is the reason for #318.

lihaibh:
Ye... ive added the 'done' handler in favor of tasks like 'clean' that use it, i assumed it will work because gulp passing this callback by default so at the time when i squashed my commits i added this change.
I forgot to run 'sanity checks' after this change and thats why it happened, thanks for the report.

PR #339, opened 2016-01-03
add logs for server startup

dwido:
It would be very helpful to know which port is running from the startup task

### DIFF
--- a/tools/utils/server.ts
+++ b/tools/utils/server.ts
@@ -1,3 +1,4 @@
+import * as util from 'gulp-util';
 import * as express from 'express';
 import * as openResource from 'open';
 import * as serveStatic from 'serve-static';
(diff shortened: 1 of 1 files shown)


mgechev:
Would you use the log function from gulp utils and capitalize server?

dwido:
Good point!

dwido:
@mgechev I have committed a change, not sure why checks are failing

alalek:
@dwido There is a merge conflict, please update your changes.
You may use `git pull --rebase` (or `git pull --rebase upstream master`) command to do this.
During conflict resolving you need to remove this line from `tools/utils/server.ts`:
```
import * as connectLivereload from 'connect-livereload';
```

mgechev:
@dwido thanks for the PR!

PR #341, opened 2016-01-04
Add Jade preprocessor.

LuxDie:
Hi! I've added support for the Jade template engine, if you'd like to feature it.


mgechev:
I don't think adding jade is necessary at this point. I'm on the same opinion about the sass support.

Probably we need to reconsider what features are essential and get rid of everything else (externalize it from the seed's core and move it to a separate examples, just like @justindujardin did in the [examples](https://github.com/mgechev/angular2-seed#examples) section).

PR #345, opened 2016-01-05
Upgraded package 'del' to v 2.x.

ouq77:
The 'del' function now returns a promise, rather than taking a callback.


mgechev:
Thanks for the PR!

PR #346, opened 2016-01-05
Switched from 'gulp-minify-css' to 'gulp-cssnano'

ouq77:
Changed from 'gulp-minify-css' to 'gulp-cssnano'.
Copy minified resources (via config.ts [is this the best way?]) from the 'tmp' location.


tools/tasks/build.html_css.prod.ts
```diff
@@ -17,7 +17,9 @@ export = function buildJSDev(gulp, plugins) {
 
     function minifyCss() {
       return gulp.src(join(APP_SRC, '**/*.css'))
-        .pipe(plugins.minifyCss())
+        .pipe(plugins.sourcemaps.init())
```
mgechev: Sourcemaps are not required for the production build.

tools/config.ts
```diff
@@ -59,7 +60,10 @@ export const NPM_DEPENDENCIES = [
 
 // Declare local files that needs to be injected
 export const APP_ASSETS = [
-  { src: `${ASSETS_SRC}/main.css`, inject: true, dest: CSS_DEST }
+  // inject minified css files from tmp
+  { src: `${TMP_ASSETS_DIR}/main.css`, inject: true, dest: CSS_DEST },
```
mgechev: This change is not required. Take a look at `[mapPath](https://github.com/mgechev/angular2-seed/blob/master/tools/tasks/build.index.ts#L30-L33).`

tools/tasks/build.html_css.prod.ts
ludohenin: I do not fully agree, I think it's fine as they are in a separate file.

mgechev: This way for each individual source map the browser has to initiate a separate HTTP request (in case they are not cached somehow).

ouq77: AFAIK source maps are only requested when the dev tools are opened. I am happy to remove it though.

tools/config.ts
ouq77: The problem is that the CSS gets minified in the 'build.html_css.prod' task and ends up in the 'tmp' assets folder. The 'build.index' then copies the unminified CSS from 'ASSETS_SRC'.

tools/config.ts
```diff
@@ -59,7 +60,8 @@ export const NPM_DEPENDENCIES = [
 
 // Declare local files that needs to be injected
 export const APP_ASSETS = [
-  { src: `${ASSETS_SRC}/main.css`, inject: true, dest: CSS_DEST }
+  // inject minified css files from tmp
+  { src: `${TMP_ASSETS_DIR}/main.css`, inject: true, dest: CSS_DEST }
```
mgechev: Is this change required given [`mapPath`](https://github.com/mgechev/angular2-seed/blob/master/tools/tasks/build.index.ts#L30-L33).

ouq77:
I will restart this on a new branch to not polute the history of config.ts and only apply the change to upgrade to gulp-cssnano.

PR #347, opened 2016-01-05
Exclude IDE generated files (.js and .js.map)

ouq77:
When using an IDE like Webstorm/IntelliJ, TypeScript files are compiled to .js and .js.map files in the same folder alongside the .ts files (by default).

Would using 'outDir' in 'tsconfig.json' be better?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,10 @@ lib
 test
 tools/typings/tsd
 tmp
+
+# Transpiled files
+gulpfile.js*
+app/**/*.js
+app/**/*.js.map
+tools/**/*.js
+tools/**/*.js.map


.gitignore
```diff
@@ -39,3 +39,10 @@ lib
 test
 tools/typings/tsd
 tmp
+
+# Transpiled files
+gulpfile.js*
+app/**/*.js
```
ouq77: Have deliberately not done '*.js*' - this also excludes '.json' files. It may or may not be an issue.

mgechev:
`outDir` seems like a better option. In some *rare* cases it may be useful to use a mixture between TypeScript and JavaScript in the project directory.

PR #348, opened 2016-01-05
feat(Protractor): add e2e testing via protractor

NathanWalker:


app/components/app/app.e2e.js
```diff
@@ -0,0 +1,19 @@
+describe('App', function() {
```
mgechev: Can we rename it to `app.e2e.ts` in order to have autocompletion? I think we already have protractor type definitions.

protractor.conf.js
```diff
@@ -0,0 +1,43 @@
+exports.config = {
+    baseUrl: 'http://localhost:5555/dist/dev',
+
+    specs: [
+        'app/**/*.e2e.js'
```
mgechev: If we change the e2e file to `ts` this should become `dist/dev/**/*.e2e.js`.

mgechev:
Great job! I added two small things.

NathanWalker:
Great suggestions @mgechev thanks! I also added more e2e tests as well to complete the coverage; they should serve as good examples of how to text various behaviors :+1:

mgechev:
@NathanWalker awesome! I'll merge tomorrow morning.

ludohenin:
Awesome

PR #354, opened 2016-01-06
Switched from 'gulp-minify-css' to 'gulp-cssnano'

ouq77:


mgechev:
LGTM, @ouq77 thanks for pointing out the problem with the location of the CSS files! I'll address it in a separate issue here.

PR #366, opened 2016-01-08
Update README.md

tarlepp:
Fix to write compiled CSS file to application destination folder and the same as the SCSS file. Generally only SCSS files should be kept in VCS.


mgechev:
Ups, thanks for the catch!

Issue #379, opened 2016-01-13
Proposal: 'organize by feature' instead of 'organize by type'


evanplaice:
Take a look at:
https://egghead.io/lessons/angular-2-organizing-angular-2-projects-by-feature

## Proposal

*Organize angular2-seed by feature instead of type*

I know it's tempting to follow the structure traditionally used by MVC platforms (ie organization by type), especially on smaller projects. I'm arguing that component-based development is a different beast and there are advantages that come with abandoning the old model.

*Example: Organization by Type*
```
- components
  - HeaderComponent
  - MainComponent
  - MarkdownEditorComponent
  - MarkdownPreviewComponent
  - FooterComponent
  - SomeOtherComponent
- services
  - SomeOtherService
  - MarkdownService
- models
  - SomeOtherModel
  - MarkdownModel
- AppComponent
```

*Example: Organize by Feature:*
```
- main
  - components
    - HeaderComponent
    - MainComponent
    - FooterComponent
- markdown
  - components   
    - MarkdownEditorComponent
    - MarkdownPreviewComponent
  - models
    - MarkdownModel
  - services  
    - MarkdownService
- shared
  - components
    - SomeOtherComponent
  - services
    - SomeOtherService
  - models
    - SomeOtherModel
- AppComponent
```

*Propositions:*

*1. Organizing by feature promotes better reusability*

Lets say I'm building an interactive Markdown editor for a site. It requires one main component and a number of smaller subcomponents, a service, and a model. The functionality is feature-complete and a use case pops up where it would be desirable to use it on another project.

With a by-type structure, reuse would require manually extracting and copying the features over into the other project. Being careful to update all of the path references, so as not to break functionality.

With a by-feature structure, simply copy the entire feature directory over because it's completely self-contained. With relative paths, nothing should break.

To take things this approach one step further. If a feature is sufficiently decoupled from the app, it becomes trivial to host the feature in it's own repo so it can be versioned independently and installed via tools like JSPM.

I understand it may seem awkward at first for devs who are used to the traditional MVC structure. But, isn't reusability the primary factor behind the shift to web components?

*2. Organizing by feature makes it easier to reason about the code*

Splitting things up by type works very well for describing the usage on small/trivial demonstrations where a component, service, model, pipe are all conveniently given the same name.

[...]

Issue #376, opened 2016-01-12
IE10 & IE11 Syntax Error: The use of a keyword for an identifier is invalid


jimulle:
When I clone the seed project and try to view it in IE10, I get the following error:
> EXCEPTION: SyntaxError: The use of a keyword for an identifier is invalid 

Unfortunately, the console output is rather cryptic and isn't providing me with much insight into the cause of this issue.

*Full Console Output:*
> SyntaxError: The use of a keyword for an identifier is invalid
   at T (http://localhost:5555/node_modules/angular2/bundles/angular2.min.js:1:1792)
   at e.prototype._createViewFactoryRuntime (http://localhost:5555/node_modules/angular2/bundles/angular2.min.js:18:4319)
   at Anonymous function (http://localhost:5555/node_modules/angular2/bundles/angular2.min.js:18:2971)
   at run (http://localhost:5555/node_modules/angular2/bundles/angular2-polyfills.js:138:7)
   at Anonymous function (http://localhost:5555/node_modules/angular2/bundles/angular2.min.js:10:9636)
   at zoneBoundFn (http://localhost:5555/node_modules/angular2/bundles/angular2-polyfills.js:111:7)
   at lib$es6$promise$$internal$$tryCatch (http://localhost:5555/node_modules/angular2/bundles/angular2-polyfills.js:1511:9)
   at lib$es6$promise$$internal$$invokeCallback (http://localhost:5555/node_modules/angular2/bundles/angular2-polyfills.js:1523:9)
   at Anonymous function (http://localhost:5555/node_modules/angular2/bundles/angular2-polyfills.js:

jimulle:
As additional clarification, this error starts after this commit
> Update to beta.1 
> 5ecfd19f2d8baacb22cc7ac907cceb4aa237d313

mgechev:
As workaround for now you can use the code under this tag https://github.com/mgechev/angular2-seed/tree/2.0.0-beta.0

jimulle:
Thanks. I actually rolled back to this commit (Merge pull request #371 from mgechev/fix-333 0a23a0398c8a762530d269a22ab28ce42a9735fc) and it works for now. I'll keep an eye out for a resolution to the issue (maybe in 2.0.0-beta.2)

mgechev:
This is weird. https://github.com/mgechev/angular2-seed/pull/371 contains only changes in the component's spec.

[...]

PR #382, opened 2016-01-14
Organize components/services/tests by feature

evanplaice:
Details and reasoning behind the changes are outlined in #379

All tests have been verified to pass prior to commit although there were some preexisting warnings.

```bash
/Users/evanplaice/Desktop/angular2-seed/app/core/components/app.ts
      line 6  col 9  unused variable: 'HTTP_PROVIDERS'
...
13 01 2016 23:36:54.562:WARN [web-server]: 404: /assets/img/smile.png
13 01 2016 23:36:54.562:WARN [web-server]: 404: /assets/img/smile.png
13 01 2016 23:36:54.562:WARN [web-server]: 404: /assets/img/smile.png
13 01 2016 23:36:54.562:WARN [web-server]: 404: /assets/img/smile.png
```


evanplaice:
I'm not sure if the `appveyor` failures could be any more cryptic if they tried to make them worse.

```bash
Failed: The use of a keyword for an identifier is invalid
...
Failed: The use of a keyword for an identifier is invalid
...
Failed: Expected ';'
```

I can't see anything obvious in the code that would cause the failures.
The first 2 seem to be related to #376.

mgechev:
I mostly like this structure. There are only two things which differ from my preferences:
- I like when specs are next to the corresponding units that we test with them
- I wouldn't name the directory that holds `AppCmp` `core`, may be `master` (which comes from the world of ASP.NET) or keep it `app`? I'm open for other naming suggestions.

evanplaice:
@mgechev 

 - I figured moving them to a separate tests folder made sense because they cover the entire component (ie components, services, and pipes). I don't see any issue with keeping them alongside the components.

- I suggest calling it 'main' instead since it's the main entry point of the application. I haven't seen 'master', although I don't do ASP.net dev. App would work but then the folder structure would be `app/app/components/app`.

What this project calls 'bootstrap.ts' I prefer to name 'app.ts'. Whereas 'main.ts' functions as the main starting point

Here's what I'd prefer:

```
- app/
  - main/
    - components/
      - footer.html
      - footer.ts      
      - header.html
      - header.ts
      - main.ts
      - main.ts
  - feature/
    - components/
    - models/
    - services/
  - shared/
    - components/
    - models/
    - services/
    - pipes/
  - app.ts
```

So the basic model is, every feature can contain it's own feature-specific `components`, `models`, `pipes`, `services`. The `shared` folder is where all the application-wide `components`, `models`, `pipes`, and `services` go.

-----

For a realistic use case, I'm planning to build a demo MarkdownEditor app. It won't require multiple pages so I'll just remove the `main` module.

```
- app/
  - markdown/
    - components/
      - markdown.ts
      - editor.ts
      - preview.ts
    - models/
      - markdown.ts
    - services/
      - markdown.ts
    - pipes/
      - md-to-html.ts
      - highlight.ts
  - app.ts
```

So, `app.ts` bootstraps the app and loads the MarkdownCmp. The MarkdownCmp imports both MarkdownEditorCmp and MarkdownPreviewCmp as views, MarkdownService to handle data storage/import/export and the pipes are used in the template to transform the data.

To reuse this, one could just copy the `markdown` feature directly into another project. It could be used to embed a complete editor, a standalone previewer, a standalone editor, or the pipes could be used standalone in an existing template. Immediate reuse requires nothing more than importing and wiring up a few modules into an existing app. That's kind of the path my ng2-markdown-component worked out (except it's just a standalone directive).

-----

With that said, naming is mostly based on subjective preference. It's probably not worth wasting a lot of time paintshedding over.

The important takeaway of this change is the preference for 'organization by feature' because I think it will scale better and maximizes reusability of Angular2 web components.

Let me know what you think and I'll make the changes.

mgechev:
Your suggestions make sense. Here is what I think according to them:
- As you said its a good idea to define a function called `main` inside `bootstrap.ts` (you call it `app.ts`). I like this! Lets rename `bootstrap.ts` -> `main.ts`. This seems more consistent to me.
- I'd also prefer to keep the `spec` and `e2e` files next to the units they test.
- Lets revert back `core` to `app.ts`, like it was initially*

*I agree that `core`, `master`, `main` or whatever is a matter of personal preference. I choose `master` because it is basically the `master` (or root) component, and all other components are its children. May be in such case `root` makes even more sense? Or we can just keep it `app` since it is the root of the application, so basically it contains it all. I think most developers will call this component differently depending on their application. So as you said loosing time in thinking of a generic name which will be later changed doesn't make a lot of sense. Lets keep it `app.ts`.

evanplaice:
OK, that should be everything.

I tried to minimize the impact of renaming bootstrap.ts to main.ts as much as possible. I wasn't sure if I should rename the `BOOTSTRAP_MODULE` in config.js.

While I was updating README.me I changed the directory structure order from by-alpha to by-kind. I think it's easier to read this way. If you don't like the switch, let me know and I'll change it back.

Sorry about commit order being out of sync. I forgot to rebase prior to pushing and had to force push to my remote after the conflicts were resolved.

mgechev:
@evanplaice looks good to me! Tomorrow or on Monday I'll merge it.

mgechev:
The final two things I found are:
- In `./tools/tasks/build.bundles.ts` change `bootstrap` to `main`
- Comment `HTTP_PROVIDERS` in `app/app/components/app.ts`, since it is unused variable and the linter throws an error

It'll be also awesome if you can squash to one commit. Great job! Looking forward to merge.

evanplaice:
Sorry about the delay. I've been busy working on other things.

The changes you requested have been made and all the commits have been squashed into one.

Bigous:
Hi @evanplaice tecnical question (a git noob one)... how did you squached the commits into one after doing the pull request and without having to make a new one?

evanplaice:
@Bigous Do an interactive rebase against another branch.

I have a remote tracking branch for upstream/master setup. So, first I updated that.
```
git remote update
```

Then I rebased against it in interactive mode
```
git checkout by-feature
git rebase -i upstream/master
```

**Step 1:**
A log will popup showing all of the commit your branch includes on top of the ref for upstrem/master. Everything starts off marked with `pick` meaning all of the commits will be maintained. To squash them change `pick` to `squash`.

**Important:** squash merges the diff of a commit into the adjacent more recent commit. If you want to merge all commits into one, always leave the most recent commit marked `pick`. I can't remember what happens if you don't, I think it's the equivalent to a soft reset where the changes are maintained but all of the commit logs are lost.

**Step 2:**

A second log will popup showing a summary log of all the commits. Basically, rearrange/rewrite the commit logs into one to outline the changes. In my case, I created a `changes` listing that basically contained the title of each commit that was squashed.

I highly recommend that the first time you attempt this, do it on a throwaway branch. This is rewriting the history in a destructive manner. Ideally, the feature branch will be backed up on a remote repo but save yourself the time and just use a throwaway branch.

PR #385, opened 2016-01-15
adding gulp plugin for typedoc

c-ice:
build.docs.ts is using this plugin but its missing from package.json


mgechev:
Would you move this change to this PR https://github.com/mgechev/angular2-seed/pull/386 ?

c-ice:
I can try it

PR #387, opened 2016-01-15
ignoring compile errors

c-ice:
This is Optional, but its drop horrible error when code is not lint free


mgechev:
I'm not sure we need this at this point. We included linting in order to show these errors.

Issue #383, opened 2016-01-14
npm run build.prod broken


tandu:
If I add `homeCmp` in `app.ts`
```typeScript
directives: [ROUTER_DIRECTIVES, HomeCmp]
```
and add home tag in `app.html`
```html
<section class="sample-app-content">
  <nav>
    <a [routerLink]="['/Home']">Home</a>
    <a [routerLink]="['/About']">About</a>
  </nav>

  <router-outlet></router-outlet>
  <home></home>
</section>
```

I get `﻿RangeError: Maximum call stack size exceeded` error when run production version of angular2-seed. It happens if I put any component into `app.html`. In my application it is `topBar` component.
<img width="1280" alt="screenshot 2016-01-14 23 31 05" src="https://cloud.githubusercontent.com/assets/273313/12336704/f895c64c-bb16-11e5-8fed-2d44960779e5.png">

mgechev:
`HomeCmp` is already added in `@RouteConfig`, you don't have to add it in `directives`.

tandu:
Yes, it solves the issue. Why it doesn't break in dev?

tandu:
I have looked at my code, I don't have component being listed in routes and have this issue.

tandu:
I have created new component `Cmp` and added it to angular2-seed. It breaks application in prod.
[angular2-seed.zip](https://github.com/mgechev/angular2-seed/files/90989/angular2-seed.zip)

mgechev:
I'll take a look at this issue tomorrow.

tandu:
Just in case it might matter - I'm testing it by running python server and then opening it in browser at `localhost:8000`:
```bash
cd dist/prod
python -m SimpleHTTPServer
```

mgechev:
@tandu the issue seems fixed in this branch https://github.com/mgechev/angular2-seed/tree/prod-enhancements.
Let me know if it solves your issue.

tandu:
I have checked, it works well, thank you.

Issue #364, opened 2016-01-08
Production build enhancements


mgechev:
- [ ] Minify CSS files https://github.com/mgechev/angular2-seed/issues/362
- [ ] Automatically set the environment variable when `gulp build.prod` is used https://github.com/mgechev/angular2-seed/issues/350
- [ ] Produce a single minified JavaScript file for all the JavaScript resources
- [ ] Use production version of Angular (the one without the template compiler) (blocked by Angular for now)

flofly:
+1 for this issue

mgechev:
@flofly you can give it a try at https://github.com/mgechev/angular2-seed/tree/prod-enhancements.

Let me know what you think.

Issue #362, opened 2016-01-08
CSS files in the prod build are not minified


mgechev:

ludohenin:
Components' css files are minified in the prod build. @mgechev Are you thinking to something else ?

mgechev:
The CSS files located at assets are not getting minified yet.

ouq77:
I've been looking at cache busting, which could take care of the minification during the revision task. Have a look here: https://github.com/mgechev/angular2-seed/compare/master...ouq77:feature/revision.

Bigous:
I thing we should copy assets directory as is, or, how could we put an example CSS file to show somewhere in the application?
A minified version of the file is not too good to be shown to the final user :D

Issue #350, opened 2016-01-05
Automatically set environment when "gulp build.prod" is used


mgechev:
Currently it is a bit annoying that when one wants to run the production build needs to explicitly set the environment:

```bash
gulp build.prod --env prod
```
instead of simply
```bash
gulp build.prod
```

ludohenin:
Once we get through #330 this issue should be soved.

PR #388, opened 2016-01-15
chore(prod): refactor the production build

mgechev:
Main changes:

- Moved the production build from `System.register` format with the `systemjs-builder` to browserify. The main purpose is to reduce the bundle size.
- Removes systemjs' config
- Creates a single bundle for all libraries and the application itself (except the shims which at this point are into a separate file, included before the app)
- Minifies all the CSS resources and bundles them into a single file
- Copies all the files different from `.ts` for `dev` and `.ts`, `.html`, `.css` for `prod` (since the templates are supposed to be inlined, the CSS files are either or bundled)
- Adds another array of shims required for the production build

Fixes https://github.com/mgechev/angular2-seed/issues/383, https://github.com/mgechev/angular2-seed/issues/364, https://github.com/mgechev/angular2-seed/issues/362, https://github.com/mgechev/angular2-seed/issues/350.


tools/utils/template_locals.ts
```diff
@@ -1,8 +1,9 @@
-import {APP_BASE, APP_DEST, APP_ROOT, APP_TITLE, SYSTEM_CONFIG, VERSION, HOT_LOADER_PORT, BOOTSTRAP_MODULE} from '../config';
+import {APP_BASE, APP_DEST, APP_ROOT, APP_TITLE, SYSTEM_CONFIG, VERSION, HOT_LOADER_PORT, BOOTSTRAP_MODULE, ENV} from '../config';
```
ludohenin: Don't you think we should `import * as CONFIG from ...` and return `CONFIG` so we never have to touch this ?

tools/config.ts
```diff
@@ -80,14 +93,7 @@ const SYSTEM_CONFIG_DEV = {
...
-  bundles: {
-    'bundles/app': ['bootstrap']
-  }
-};
-
-export const SYSTEM_CONFIG = ENV === 'dev' ? SYSTEM_CONFIG_DEV : SYSTEM_CONFIG_PROD;
+export const SYSTEM_CONFIG = SYSTEM_CONFIG_DEV;
 
 // This is important to keep clean module names as 'module name == module uri'.
 export const SYSTEM_CONFIG_BUILDER = {
```
ludohenin: I guess this is not required anymore.

tools/tasks/build.assets.dev.ts
```diff
@@ -4,12 +4,8 @@ import {APP_SRC, APP_DEST} from '../config';
 export = function buildImagesDev(gulp, plugins) {
```
ludohenin: function names isn't correct :)

tools/tasks/build.assets.prod.ts
```diff
@@ -2,13 +2,25 @@ import {join} from 'path';
 import {APP_SRC, APP_DEST} from '../config';
 
 export = function buildImagesDev(gulp, plugins) {
```
ludohenin: function names isn't correct :)

tools/config.ts
```diff
@@ -29,44 +34,52 @@ export const TEST_DEST            = 'test';
...
+  { src: 'systemjs/dist/system.src.js', inject: 'shims', dest: JS_DEST },
+  { src: 'angular2/bundles/angular2-polyfills.js', inject: 'shims', dest: JS_DEST },
+  { src: 'rxjs/bundles/Rx.js', inject: 'libs', dest: JS_DEST },
+  { src: 'angular2/bundles/angular2.js', inject: 'libs', dest: JS_DEST },
+  { src: 'angular2/bundles/router.js', inject: 'libs', dest: JS_DEST },
+  { src: 'angular2/bundles/http.js', inject: 'libs', dest: JS_DEST },
+  { src: 'bootstrap/dist/css/bootstrap.css', inject: true, dest: CSS_DEST }
+]);
+
+export const PROD_NPM_DEPENDENCIES: InjectableDependency[] = normalizeDependencies([
```
ludohenin: `angular2-polyfills.js` is missing here (zone not included, so change detection doesn't work in about page)

tools/utils/template_locals.ts
mgechev: Yep, makes sense.

tools/config.ts
```diff
@@ -29,44 +34,53 @@ export const TEST_DEST            = 'test';
...
+  { src: 'reflect-metadata/Reflect.js', inject: 'shims' },
+  { src: 'es6-shim/es6-shim.min.js', inject: 'shims' },
+  { src: 'angular2/bundles/angular2-polyfills.min.js', inject: 'libs' },
```
mgechev: @ludohenin thanks for the catch! I added it with the latest update.

PR #420, opened 2016-01-30
Update to beta.2

mgechev:
Changes:
- Update the angular2 and angular2-hot-loader packages versions.
- Update the tests according to the latest [breaking changes](https://github.com/angular/angular/blob/master/CHANGELOG.md#breaking-changes).
- Update the SystemJS config in order to speedup the dev build page load time.
- Merged with https://github.com/mgechev/angular2-seed/pull/411.

PR #424, opened 2016-02-01
fix: allow to name bootstrap file differently than main.ts

juristr:
Hi,

I just migrated a simple demo project of mine to use this angular2-seed project. I had some trouble, though, because my startup file was named `bootstrap.ts` rather than `main.ts`.

By looking at the `config.ts` I found there's a variable `BOOTSTRAP_MODULE`. I changed it like this:

```javascript
export const BOOTSTRAP_MODULE     = ENABLE_HOT_LOADING ? 'hot_loader_main' : 'bootstrap';
```
..but it didn't work. After digging a bit in the configuration, I found that the reason is `main` being hardcoded in a couple of places. This PR fixes this. Obviously this shouldn't be a big deal as ppl could only rename their bootstrap file and it's done :smile:.

The only thing I don't really like is that readability of the `SYSTEM_CONFIG_DEV` in `config.ts` suffers a bit. Not sure if there's a better option.. :smile:

### DIFF
--- a/tools/config.ts
+++ b/tools/config.ts
@@ -88,11 +88,11 @@ export const PROD_DEPENDENCIES = PROD_NPM_DEPENDENCIES.concat(APP_ASSETS);
 const SYSTEM_CONFIG_DEV = {
   defaultJSExtensions: true,
   paths: {
-    'main': `${APP_ROOT}main`,
     'hot_loader_main': `${APP_ROOT}hot_loader_main`,
     '*': `${APP_BASE}node_modules/*`
   }
 };
+SYSTEM_CONFIG_DEV.paths[BOOTSTRAP_MODULE] = `${APP_ROOT}${BOOTSTRAP_MODULE}`;
 
 export const SYSTEM_CONFIG = SYSTEM_CONFIG_DEV;
 
--- a/tools/tasks/build.bundles.ts
+++ b/tools/tasks/build.bundles.ts
@@ -2,6 +2,7 @@ import * as merge from 'merge-stream';
 import {join} from 'path';
 import * as browserify from 'browserify';
 import {
+  BOOTSTRAP_MODULE,
   PROD_DEPENDENCIES,
   JS_PROD_SHIMS_BUNDLE,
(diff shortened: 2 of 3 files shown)


NathanWalker:
Thanks for this @juristr . I also ended up needing to do the same thing when extending here:
https://github.com/NathanWalker/angular2-seed-advanced/blob/development/tools/config.ts#L107

There, I also found it beneficial to make the bootstrap module a command line argument for different runtime configuration targets. I believe something like this would be good in this seed.

Bigous:
@NathanWalker I think you should merge advanced back here :D

NathanWalker:
Lol very funny! In all seriousness there's certain things that will make
sense here for most baseline projects and other large scale projects could
look into advanced seed.

juristr:
@NathanWalker I just don't get why the tests fail on AppVeyor? Was this introduced by my PR or was it already before?

NathanWalker:
It will be fixed once this PR goes in:
https://github.com/mgechev/angular2-seed/pull/420

juristr:
oh I see :+1:

mgechev:
@juristr thanks for the catch!

tools/config.ts
```diff
@@ -88,11 +88,11 @@ export const PROD_DEPENDENCIES = PROD_NPM_DEPENDENCIES.concat(APP_ASSETS);
 const SYSTEM_CONFIG_DEV = {
   defaultJSExtensions: true,
   paths: {
-    'main': `${APP_ROOT}main`,
```
ludohenin: You can use ES2015 computed property name feature here I guess.
`[BOOTSTRAP_MODULE]: '${APP_ROOT}${BOOTSTRAP_MODULE}'`

juristr: Oh, I was searching for it as I didn't remember the syntax. Knew I saw something like this somewhere. Would definitely help with readability.

I'll send an update later, thx :+1:

ludohenin: This has in fact changed in another PR. There is nothing to update.

Issue #445, opened 2016-02-07
e2e tests have stop working after moving to / as the base URL


nareshbhatia:
End-to-End tests are still using /dev/dist as the baseUrl. Also the tests themselves are using absolute paths, so Protractor's baseUrl configuration is being ignored. I will send a PR to fix this.

nareshbhatia:
Fixed with PR #446

PR #446, opened 2016-02-07
Fixed URLs for e2e tests

nareshbhatia:
Issue #445


mgechev:
Thanks for the catch!

Issue #447, opened 2016-02-07
Inconsistency in test compilation


nareshbhatia:
While unit tests are compiled to `/test`, e2e tests are compiled to `/dist/dev`. This has the following downsides:
- Unit tests and e2e tests are treated inconsistently.
- `/dist/dev` has a mix of application js and e2e js.
- Many people like to keep test source in `/test`, but this name is now taken to store unit test compiles.

Can I suggest the following organization:
- Unit tests get compiled to `/dist/test/spec`
- e2e tests get compiled to `/dist/test/e2e`

This not only makes the compilation of unit and e2e tests consistent, but also frees up the `/test` directory for people wishing to use it for storing test source.

I can create a PR if this seems like a good idea.

P.S. I personally don't like to mix application and test source. I can understand why some people do, perhaps to keep it together with the source. I tend to create far more test files than application files, hence mixing them obscures the application source. Also I tend to write tests in layers, following the [Test Automation Pyramid](https://www.mountaingoatsoftware.com/blog/the-forgotten-layer-of-the-test-automation-pyramid) suggested by Mike Cohn. I find it easier to do this if my tests are in a separate directory structure. In any case, it is a matter of opinion and I would appreciate if the seed project accommodates both ways of working.

mgechev:
Seems like a very good idea to me and we'll really appreciate your help!

nareshbhatia:
Sounds good. Will start working on it today.

mgechev:
Great! Looking forward for the PR!

mgechev:
Fixed by https://github.com/mgechev/angular2-seed/pull/476.

nareshbhatia:
@mgechev, I am trying to extend the build to include Sass and realizing that the decision to split the dev, unit and e2e builds into 3 different directories may not be a good one. Since Sass requires an additional step of converting from .scss to .css, I am finding that I have to write additional tasks for each type of build. I am now thinking that it would be better to use `/dist/dev` as a *playground* for all dev activities, including unit and e2e testing. It will make the build process much simpler.

The only question in my mind is why was the e2e build in a separate directory earlier (`/test`)? As far as I can tell, it can be folded quite easily into `/dist/dev`. Am I missing something? If not, I can work on a PR.

mgechev:
@nareshbhatia I don't think there's any particular reason. @NathanWalker added the e2e testing support so he can verify.

ludohenin:
[...]

PR #476, opened 2016-02-13
Fix test compilation inconsistency

nareshbhatia:
Fixes issue #447 

- Unit tests get compiled to `/dist/test/spec`
- E2e tests get compiled to `/dist/test/e2e`
- Added `TEST_SRC` constant to `config.js`. By default it points to `src`, but allows the possibility to make it point to `test` (thus keeping app source and test source in different folders).


mgechev:
Great job! I'll run it and merge it today.

PR #494, opened 2016-02-16
Strip first two directories instead of keeping the last two

Groschenroman:
Since we don't know how many subdirectories there will be, we'ld rather strip the leading `/dist/dev` directories instead of keeping the last two. 

**Note:** Even though we only want to get rid of _two_ directories, the starting index needs to be `3` because the first forward slash is its own value.


mgechev:
@Groschenroman makes sense. Would you rebase and push again?

Groschenroman:
Something went wrong with the rebasing (it was challenging to get it anyway…).
I'll close this PR and open a new one.

PR #498, opened 2016-02-16
Strip first two directories instead of keeping the last two

Groschenroman:
I was troubled trying to rebase the first PR. This should work fine now, I hope.


mgechev:
Thanks for the PR!

PR #501, opened 2016-02-16
Cleanup: Cleanup npm from npm

jesperronn:
During postinstall we start gulp in order to start npm and run `npm prune`

We also clean the dist folder, which is completely irrelevant during install.

I suggest this to change so that the postinstall step directly prunes unused npm
packages.

Since we remove the extra gulp step, it can also shave 2-3 seconds off of a CI
build step.


mgechev:
Great one!

PR #503, opened 2016-02-16
fix(build): npm run reinstall 

### LINKED ISSUE #474: "npm run reinstall" fails
Hi. Running `npm run reinstall` exits with error "The system cannot find the path specified." **after** deleting the `node_modules` directory.
I'm running Windows 10, Node.js v5.3.0 and Npm v3.5.2.

d3viant0ne:
> npm run reinstall fails on Win10 on the rimraf command. When you dig far enough you end up with an ENOTEMPTY issue in rimraf

- Bumped minimum version of rimraf to 2.5.2
- Added wildcard to folder path to working around weird windows semantics
- No cleans cache before removing node_modules directory

Fix verified on WIndows 10 without regression on OSX 10.11.3 or Ubuntu 14.04.03

resolves #474


mgechev:
Awesome job!

PR #507, opened 2016-02-17
(performance) Faster loading of tasks (skip unused)

jesperronn:
This has annoyed me forever. On my machine, calling the custom "task()" method takes roughly half a
second.

Depending on which gulp task you execute, there is no reason to load these tasks in the top of
the gulpfile. I wrap them in a function, which hides the underlying true problem.

The real problem is of course that the loading of tasks take a long time, and hides unused tasks so that we never delete them.

But that will be a separate problem requiring a bigger fix.

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -3,13 +3,12 @@ import {runSequence, task} from './tools/utils';
 
 // --------------
 // Clean (override).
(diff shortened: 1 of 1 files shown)


mgechev:
@jesperronn nice one! I'll take a look at merge today.

PR #513, opened 2016-02-18
chore: add angular 2 specific static code validation

mgechev:
I'm working on a [static code analyzer](http://blog.mgechev.com/2016/02/18/angular2-style-guide-best-practices-lint-tslint) which verifies that given project follows the best practices described in the [Angular 2 Style Guide](https://github.com/mgechev/angular2-style-guide).

It will be configurable so later if any other style guide is released one be able to adjust the configuration options based on his own preferences.


tools/tasks/tslint.ts
```diff
@@ -11,7 +11,9 @@ export = function tslint(gulp, plugins) {
     ];
 
     return gulp.src(src)
-      .pipe(plugins.tslint())
+      .pipe(plugins.tslint({
+        rulesDirectory: NG2LINT_RULES
```
ludohenin: Maybe that one should be included in `tsconfig` instead to leverage linting straight from the IDE.

mgechev: Yes, I intended to do so but I didn't manage to make it work. Not sure if gulp-tslint recognizes this option.

ludohenin: Yep. For VSCode it should actually be added to `.vscode/settings.json` (https://marketplace.visualstudio.com/items?itemName=eg2.tslint)

ludohenin: It could still be added to `tslin.json` so if running `tslint` in a shell it will work as well.

mgechev: Yep, definitely adding it to `tslint.json` is a good idea.

PR #524, opened 2016-02-21
chore: Update gulp-inline-ng2-template

ludohenin:
[gulp-inline-ng2-template](https://github.com/ludohenin/gulp-inline-ng2-template) now allow to you to process templates and styles.


mgechev:
I guess the `rulesDirectory` needs to be specified. It is strange that AppVeyor didn't complain before.

Issue #506, opened 2016-02-17
Workflow for feature forks


nareshbhatia:
I am about to create a fork which will be adding Sass and specific component frameworks. Looking for ideas/workflows which will allow me to keep in sync with mgechev/angular2-seed without undoing my feature specific changes.

mgechev:
@NathanWalker is doing something similar in his [angular2-seed-advanced](https://github.com/NathanWalker/angular2-seed-advanced).

nareshbhatia:
Agree, seems like a hard problem to solve with Git. I see that @NathanWalker's project is not a fork of angular2-seed, although he started with the angular2-seed as the base. Love to hear more about his approach.

d3viant0ne:
@nareshbhatia - There is also discussion around adding support for pre-processors to inlineNg2Template. Seems a reasonable way to allow people to use sass, scss, stylus & so on while simplifying maintenance of the core.

nareshbhatia:
Agree. I am banking on @ludohenin to provide that feature (see issue #447). However, I will still have to create a fork for other integrations and keep it in sync with mgechev/angular2-seed to preserve all the goodness :-).

ludohenin:
Regarding the seed update, for me best option would be to move the tools to an npm package (say `angular2-seed-tools`) which would be overridden by the project specific config and tasks. This way we can keep tagging the the seed as new ng2 releases come along and version the tools separately.

d3viant0ne:
Certainly appears to be the only real seamless way to keep developing the seed without stepping on peoples toes.

NathanWalker:
@ludohenin that's an excellent idea. 

@nareshbhatia Problem with `feature forks` is that it rolls all issues into the parent, which is a problem. You wouldn't want all issues with your feature fork going back and spamming the parent. Which is why I chose **not** to make it a fork. Best way I've found is to:
* download repo as a zip file
* setup remote origin as your own repo
* setup remote upstream as `angular2-seed`
* periodically merge upstream changes into your "fork"

This works alright but obviously creates conflicts and depending on how far you ventured off the parent, could be really bad. Just depends in your case how much you are planning on changing things.
Ideally, what @ludohenin mentioned would make this sort of thing ideal. :+1:

nareshbhatia:
@NathanWalker, hmm that's an interesting one - issues from the feature fork roll up to parent!!!! That's not very good. I like your workaround - will go with that.

[...]

PR #528, opened 2016-02-22
refactor(tools): Update folder structure

ludohenin:
To follow up on issue #506 and seed update & extensibility here is a proposal for a migration path starting with a new structure for the seed tools.

There is no breaking changes at the moment, everything still work as of now.

After a deeper thought, packaging the tools doesn't seems to be ideal.
* another level of complexity
* doesn't solve the "how do I extend it?"

Whereas with this structure we give ourself an easy path to just update the seed files but never the project ones (or any other - company, customer ...) and extend from the seed.

Next steps could be:
* [x] compile `./tools/**/*.ts` only 
  * we should see a significant performance improvement
  * easier debugging of gulp tasks
* [ ] manage dependencies and typings
  * ideally it would be nice not to mess up `app` deps and `tools` deps
* [x] define interfaces for the utils (see `./tools/utils.ts`)
* [x] define a better general tools API (especially tasks referencing/loading)
* [x] migrate tasks
* [ ] add utilities to seamlessly update the seed files

any thoughts ?


NathanWalker:
I really like where you're going with this @ludohenin nice work.

d3viant0ne:
@ludohenin - Given the conversation this afternoon, compiling the tools seems to be a needed change.
I'm liking this idea more than packaging now that I have given it more thought. 

Short of delivering patch files, i'm curious what tools you are thinking about adding that would update seed files? Most of the end user repos I have seen have modified tooling.

ludohenin:
I've added compilation of `tools/**/*.ts` and `gulpfile.ts`.
The perfs improvement isn't big using js over ts files
* `build.dev`
  * JS: ~ 4.5s
  * TS: ~ 7s
* `build.prod`
  * JS: ~ 17s
  * TS: ~ 22s

Any way it's an opt-in, we can use either ones.

I've added task debugging support for VS Code
__.vscode/launch.json__
```json
{
	"version": "0.2.0",
	"configurations": [
		{
			"name": "Debug gulp task",
			"type": "node",
			"request": "launch",
			"program": "${workspaceRoot}/tools/debug.js",
			"stopOnEntry": false,
			"args": ["--task", "build.dev"],         <= change the task name here
			"cwd": "${workspaceRoot}",
			"runtimeExecutable": null,
			"runtimeArgs": [
				"--nolazy"
			],
			"env": {
				"NODE_ENV": "development"
			},
			"externalConsole": true,
			"sourceMaps": true,
			"outDir": null
		}
	]
}
```

mgechev:
Awesome progress!

ludohenin:
So to continue further on this, my proposal for the pending points

### Tools API

* Task loading:
  * loads tasks from different directories with override for identical file (task) names (the last wins)
  ```es6
  // In gupfile.ts
  loadTasks(PATH_TO_SEED_TASKS);
  loadTasks(PATH_TO_PROJECT_TASKS);
  loadTasks(/* wherever */);
  ```
  * Register all tasks found so we don't need to define tasks as we are currently doing while still _requiring_ tasks on demand (ts perf).
  ```es6
  // Drop that kind of thing, will be available in the gulp registry
  gulp.task('serve.coverage', () => task('serve.coverage'));
  ```

* drop the `task` function as it is represented
  ```es6
  // Drop this, and make a different task for each (can rely upon a utility function to keep things DRY)
  gulp.task('clean', done => task('clean', 'all')(done));
  gulp.task('clean.dev', done => task('clean', 'dev')(done));
  gulp.task('clean.prod', done => task('clean', 'prod')(done));
  gulp.task('clean.tools', done => task('clean', 'tools')(done));
  ```

* Refactor task 
 ```es6
  // Drop the current function signature ...
  export = function task(gulp, plugins, option) {
    return done => { /* Task */ }
  }

  // ... in favor of something simpler (import whatever's needed for the task to run)
  export = done => { /* Task */ }
  // or
  export = () => { /* Task */ }
  ```

* Improve `runSequence` abstraction to support nested arrays (better perf by having parallel tasks execution)

### Utilis interface

This is what we have 
```ts
export interface ISeedUtils {
  // code_change_tools
  listen: () => void;
  changesd: (MSFileSaver: string|string[]) => void;

  // server
  serveSPA: () => void;
  notifyLiveReload: (e: any) => void;
  serveDocs: () => void;
  serveCoverage: () => void;

  // tasks_tools
  loadTasks: (paths: string[]) => void;                             // To come
  task: (taskname: string, options?: any) => ITask;        // To be dropped
  runSequence: (...sequence: any[]) => (...sequence: any[]) => any;

  // template_locals
  templateLocals: () => any;

  tsProjectFn: (Plugin: any) => any;
}

interface ICallback {
  (): void;
}

interface ITask {
  (done?: ICallback): void|NodeJS.ReadWriteStream;
}
```

### Tasks Migration

Will start if you agree on the proposed changes

### App and tools dependencies

I don't see any good way to do that. Any ideas ?

### Seed update

Utility/CLI tools to update the seed files using something like `nodegit` would have to come afterward.

mgechev:
All proposed changes sound good to me. I'd only prefer keeping the `utils` module and having four interfaces:

```ts
// Personally don't like the name here but I wasn't able to come up with something better
export interface ISeedCodeChangeUtils {
  // code_change_tools
  listen: () => void;
  changesd: (MSFileSaver: string|string[]) => void;
};

export interface ISeedServerUtils {
  // server
  serveSPA: () => void;
  notifyLiveReload: (e: any) => void;
  serveDocs: () => void;
  serveCoverage: () => void;
};

export interface ISeedTasksUtils {
  // tasks_tools
  loadTasks: (paths: string[]) => void;                             // To come
  task: (taskname: string, options?: any) => ITask;        // To be dropped
  runSequence: (...sequence: any[]) => (...sequence: any[]) => any;
};

export interface ISeedTemplateLocals {
  // template_locals
  templateLocals: () => any;

  tsProjectFn: (Plugin: any) => any;
};
```

This follows the interface segregation principle better and in long term should be easier to maintain.

ludohenin:
I think it's there now. Hope I've not messed the things up too much with the changes on the master branch.

I changed my mind about:
* The utils, they're not merge-able (as initially thought)
* I dropped the config interface and made them classes, a lot easier to maintain and extend (intellisence in vscode is broken, but [it should be fixed](https://github.com/Microsoft/vscode/issues/3515))

ludohenin:
And I'll have to document this a bit ... but it doesn't prevent to merging it.

mgechev:
@ludohenin awesome job, really! I'll review it and merge it tomorrow.

d3viant0ne:
@ludohenin - Really digging this as I'm getting into it more this morning. Going to make life much easier for people building on top of the seed.

No to prepare for the onslaught of questions we are all going to get over the next week :+1:

ludohenin:
Yeah!

PR #536, opened 2016-02-25
Cleanup

ludohenin:
Just a bit of cleanup (remove dead code and useless module).
`dest` prop of `IDependency` has became obsolete.
I'll update the wiki once merged.


ludohenin:
@robbatt You were right, there's a mistake. `ASSETS_SRC` must be used instead of `APP_ASSETS`.
Will fix in a minute

PR #542, opened 2016-02-27
Add base tag to index.html

LuxDie:
Add base tag so navigation won't break when the app base is other than the root domain.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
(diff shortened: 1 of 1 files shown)


ludohenin:
Wasn't this (https://github.com/mgechev/angular2-seed/blob/master/src/main.ts#L10) enough ?

mgechev:
This will help for loading resources such as js, css files and images.

ludohenin:
I'll clean the tools/utils up as it's quite inconsistent the way we handle this.

Issue #548, opened 2016-02-28
SyntaxError: Expected ')' in Prod build in IE 11


NareshPagadala:
I have downloaded the structure and created the PROD build and deployed the code.It was working fine in Mozilla and Firefox. but in IE11, we are getting a strange error --> SyntaxError: Expected ')'.

When I changed the value of minify to false in the file build.bundles.app.ts , its working fine in IE as well. Which means its causing a break when we are trying to minify the js.

Before:

const BUNDLER_OPTIONS = {
  format: 'cjs',
  minify: true,
  mangle: false
};

After:

const BUNDLER_OPTIONS = {
  format: 'cjs',
  minify: false,
  mangle: false
};

Can this be resolved as without minifying the js, its taking time to load the js files.

Screenshot of the error attached.
![ie_error_afterminify](https://cloud.githubusercontent.com/assets/17527026/13379230/edea68cc-de45-11e5-9382-526c897b357a.png)

d3viant0ne:
@NareshPagadala - https://github.com/mgechev/angular2-seed/issues/510

Issue #510, opened 2016-02-18
IE: Issue with app.js bundle, when minify is set to "true" for systemjs builder


dimarzionist:
When opening production dist in IE11, you get syntax error in app.js.
Appears after migration from browserify to sustemjs builder.
Only happens when "minify" set to true for bundler options.

Please see the attachment for full snapshot of error log console from IE. 
IE11,
npm 3.7.2
node 5.3.0

EXCEPTION: SyntaxError: Expected ')'
   EXCEPTION: SyntaxError: Expected ')'
   STACKTRACE:
   SyntaxError: Expected ')'
   at evalExpression (file:///C:/dev/git/angular2-seed/dist/prod/js/app.js?1455752277346:27:30172)
   at TemplateCompiler.prototype._createViewFactoryRuntime (file:///C:/dev/git/angular2-seed/dist/prod/js/app.js?1455752277346:8:20302)
   at Anonymous function (file:///C:/dev/git/angular2-seed/dist/prod/js/app.js?1455752277346:8:17949)
...

![capture](https://cloud.githubusercontent.com/assets/631258/13129128/e0b3cf50-d62e-11e5-89ef-9f665e5d6a66.PNG)

Bigous:
Just to add some information... on Microsoft Edge (aka IE 12) the problem does not occurs.

d3viant0ne:
@dimarzionist I'll spin up a windows vm and take a look at this

d3viant0ne:
@Bigous Can you tag this one, That's confirmed on Win10/IE 11

d3viant0ne:
I'll see if I can run down the "why" portion

dimarzionist:
Could be related to this:
https://github.com/angular/angular/issues/6695

d3viant0ne:
It is the compression version of the problem. it works if you just mangle it.

Bigous:
Removed the bug label - @dimarzionist is right.

Bigous:
Mangle the code can cause another issues in injector according to the same issue report.

PR #552, opened 2016-02-29
(fix): IE shims patch

markharding:
This is a temporary patch to resolve issues with IE browsers. See #510 #548 and https://github.com/angular/angular/issues/7144


mgechev:
Thanks for the PR!

tools/config.ts
```diff
@@ -66,6 +66,7 @@ export const DEV_NPM_DEPENDENCIES: IDependency[] = normalizeDependencies([
   { src: 'reflect-metadata/Reflect.js', inject: 'shims' },
```
js02sixty: T

jesperronn:
@mgechev and @markharding wouldn't we have to add the shim to the loaded test files in karma.conf.js or test-main.js  as well? If not, there will still be IE testing failures from appveyor, right?

PR #555, opened 2016-02-29
fix: browserSync server routes with APP_BASE

njs50:
fixes the routes when --base is specified on server start

### DIFF
--- a/tools/utils/code_change_tools.ts
+++ b/tools/utils/code_change_tools.ts
@@ -1,17 +1,24 @@
-import {PORT, APP_DEST} from '../config';
+import {PORT, APP_DEST, APP_BASE, DIST_DIR} from '../config';
 import * as browserSync from 'browser-sync';
 
 let runServer = () => {
+  let baseDir = APP_DEST;
   let routes:any = {
-    [`/${APP_DEST}`]: APP_DEST,
-    '/node_modules': 'node_modules'
+    [`${APP_BASE}${APP_DEST}`]: APP_DEST,
+    [`${APP_BASE}node_modules`]: 'node_modules',
   };
+
+  if (APP_BASE !== '/') {
+    routes[`${APP_BASE}`] = APP_DEST;
+    baseDir = `${DIST_DIR}/empty/`;
+  }
+
   browserSync({
-    middleware: [require('connect-history-api-fallback')()],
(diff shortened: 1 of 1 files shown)


njs50:
i dunno if there is some better way to set that up, but this at least makes it work. 

It might be nicer if the server baseDir was empty, instead of having something which isn't really correct when --base is specified. 

I think the only way to do that is make an empty dir someplace and set that to the base, which isn't nice either :-/

njs50:
I updated the PR so that now it won't serve any files under / when --base is specified, also fixes the connect-history-api-fallback setup so it loads the index.html from the base dir.

I'm not sure about if setting the base dir to dist/empty (non existent) is bad, but it doesn't seem to cause any issues and seems better than serving things which shouldn't be there.

mgechev:
Thanks! I'll take a look at the PR tomorrow.

njs50:
no prob, let me know if theres any updates you'd like me to make. 

we use gh-pages for some of our docs and it's handy to have the same urls when running locally :-)

tools/utils/code_change_tools.ts
```diff
@@ -1,17 +1,24 @@
...
   let routes:any = {
-    [`/${APP_DEST}`]: APP_DEST,
-    '/node_modules': 'node_modules'
+    [`${APP_BASE}${APP_DEST}`]: APP_DEST,
+    [`${APP_BASE}node_modules`]: 'node_modules',
   };
+
+  if (APP_BASE !== '/') {
+    routes[`${APP_BASE}`] = APP_DEST;
+    baseDir = `${DIST_DIR}/empty/`;
```
ludohenin: @njs50 don't understand why this line ... removing it doesn't affect anything.

njs50: @ludohenin: removing it shouldn't stop anything working, but removing it does make things work that shouldn't. 

If it's not there then the project files are served at both / and /APP_BASE/, which is why the connect-history-api-fallback was working when it shouldn't of been. i.e it was loading /index.html instead of /APP_BASE/index.html. 

It's a problem if you want your dev environment to work the same way as production.

ludohenin: ok, thanks

njs50: there might be a nicer way to do it, but as far as I can tell browserSync needs to have a baseDir so the options seem somewhat limited.

Issue #556, opened 2016-03-01
Extend css task, add css-lint task


tsm91:
I was thinking about implementing the following css and css-lint tasks, at least for myself. I know the project is very barebones so i ask before i even consider open a PR if any of these are welcome in the code / wiki page.

For `minifyComponentCss` and `processExternalCss` tasks:

```js

/**
 * how it works:
 * 1, get all external css file from `APP_ASSETS`
 * 2, minify them
 * 3, concatenate them to app.css
 * 4, split it out to `${DIST_DIR}/${ENV}/css`
 *
 * In this case it handles only /src/assets/main.css.
 *
 * TODO:
 * 1, use cssnano as a PostCSS plugin
 * 2, add autoprefixer
 * 3, add postcss-assets
 * 4, optimize for incremental build ( gulp-cached )
 *
 * optional TODO (for wiki page, because it should be major** preprocessor agnostic):
 * 1, use sass
 * 2, add sourcemaps
 * 3, optimize for incremental build ( gulp-progeny )
 *
 */
```

Add a new `css-lint` task which is going to do the following:

```js
/**
 *
 * A global css linter
 *
 * 1, use stylelint, doiuse, colorguard
 * 2, optimize for incremental build ( gulp-cached )
 * 3, filter out vendor css with postcss-filter-stream**
 *
 */
```

1**, Under major preprocessors i mean sass, less, stylus and precss. You can add any of these to the postcss stack without breaking it.

2**, For this filter the `IDependency` interface needs to be extended with `vendor` property. Because we don't want to see lint errors from third party libraries.

So again, i will implement these anyways for personal use. I just want to contribute if this is needed in any form (in the wiki, in the code).

the-ult:
Sounds interesting. Like the idea of splitting the build and linting in 2 tasks.
We are trying/playing with `postcss` as well. And are removing `sass` completely from our build.

Still trying to get the css working per component.
This is what we have so far (WIP):
```typescript
import {join} from 'path';
import {APP_SRC, CSS_DEST} from '../config';
// import {cssnext} from 'cssnext';


export = function buildPostcssDev(gulp, plugins, option) {
  return function() {
    let atImport = require('postcss-import');
    let cssnext = require('postcss-cssnext');
    let lost = require('lost');
    // let autoprefixer = require('autoprefixer');
    // let doiuse = require('doiuse');

    const processors = [
      atImport(),
      // add your 'plugins' here
[...]

PR #562, opened 2016-03-01
add css lint task

tsm91:
As discussed here #556. This is the css lint task implementation. I messed up and created #561 from the master branch, so this includes all of #561's changes. That one should be merged / rejected first.

linters:
* stylelint
* doiuse
* colorguard


gulpfile.ts
```diff
@@ -43,8 +45,10 @@ gulp.task('build.e2e', done =>
 gulp.task('build.prod', done =>
   runSequence('clean.prod',
               'tslint',
+              'css-lint',
```
mgechev: Would you name the file with `_` instead if dash or better remove the dash at all, in order to follow the convention in the rest of the tasks.

mgechev:
LGTM, great job!

Do you think we can make csslint make sure the project is following [Google's style guide](https://google.github.io/styleguide/htmlcssguide.xml)?

mgechev:
https://github.com/mgechev/angular2-seed/pull/561 seems part of this PR. Lets close it.

gulpfile.ts
tsm91: done

mgechev:
@tsm91 what do you think about the Google style guide? Are there any ready-to-use configuration files? If not, we can stick to the current rules we already have.

tsm91:
> Do you think we can make csslint make sure the project is following Google's style guide?

I think so. Would make sense to use the google css guidelines. In fact the [stylelint-config-standard](https://github.com/stylelint/stylelint-config-standard#stylelint-config-standard) uses some part of Google's style guides. 

Unfortunately i haven't found a stylelint rule implementation of it. I am sure it is possible but it would take a lot of work.

> #561 seems part of this PR. Lets close it.

Yeah sorry, i was working in the master branch instead of creating a separate one -,-

-----

1, The [postcss-assets](https://github.com/assetsjs/postcss-assets) should be mentioned somewhere in the readme to avoid confusion and point out the URL resolution possibilities.

2, The [Add-custom-Gulp-task
](https://github.com/mgechev/angular2-seed/wiki/Add-custom-Gulp-task) wiki page should also be updated, to add sass and source maps. Since you only need to drop 3 `.pipe` to build.html_css.ts it should be another wiki article. And the add custom task article should be changed to add some other rudimentary task.

mgechev:
I'm wondering if postcss-assets will have more negative rather than a positive impact on the seed.

In general I find it very useful and I'd include it in my project. On the other hand, it brings one more level of indirection and may be confusing for the developers who are not familiar with it.

// cc @ludohenin @nareshbhatia @Bigous

tsm91:
Correct, it is certainly possible. In fact it is very handy to just write `resolve()` to font-face and background urls and it will resolve the file no matter where it is inside of `/src/assets`. 

I think it is very easy to pick it up, also this is an opt-in feature so it is not necessary to use it. A few lines of comment right above `background: resolve("more.svg");` would clear the confusion i think.

ludohenin:
I'm not familiar with `postcss` stuffs so difficult to say. If it's an opt-in then that's probably fine.

d3viant0ne:
Postcss takes a bit of mental remapping compared to most of the current tools. That said, it's powerful / fast and pluggable which makes it attractive. I've only tinkered with it & I have thus far been impressed.

IMO - So long as it is opt-in and well documented, I think it has a place here. It's rapidly gained in popularity and the plugin ecosystem seems to double every time I go look.

mgechev:
@tsm91 a rebase will be required. There are a lot of changes because of the latest refactoring which aims bigger modularity of the seed so resolving the merge conflicts will not be trivial...

If you want you can open a new PR. about `postcss-assets`, how about not including them in the build process but adding a sample integration page in the wiki? How does it sound?

d3viant0ne:
@tsm91 @mgechev - I agree with the sample integration page idea for `postcss-assets` keeps things nice and pre/post processor agnostic.

@tsm91 - As a side note, having just done this with a pre-refactoring branch. It was significantly faster to create a new branch and recommit the changes than it was to try and work them into a rebase.

mgechev:
@tsm91 any updates?

tsm91:
@d3viant0ne @mgechev hey guys, sorry i am in a middle of a moving process

had to paint the walls of the flat, package up my stuff, but i will do this today and re-submit the PR

i will close this one and open a new one, the re-base process seem to be scary :D

mgechev:
No worries. There were some huge changes in the seed so I agree that a new PR will be an easier option. Looking forward for it! Btw, lets not include `postcss-assets` initially. I'll be happy if we have a wiki article for it (since I'll be one of the users of it) but the level of indirection it brings could be a bit frustrating.

tsm91:
@mgechev roger that, i wont include `postcss-assets` only as a wiki tutorial

mgechev:
@tsm91 sounds awesome! Looking forward for it!

sullyme:
`css-lint` disappeared?

PR #565, opened 2016-03-02
fix: karma watch

### LINKED ISSUE #551: karma not in sync with build.test.watch on Windows
We're having problems with karma re-running before build.test has completed. We initially thought it was our 'fork' but tested it on the seed and get the same behavior.

mgechev:
Fixes #551


ludohenin:
The problem with a single shell is that you can't debug in chrome (a new window is opened/closed on every changes).
Kama is also not started when first ran, only after a first change is made.

mgechev:
That is a good point. Lets see the outcome of #551.

PR #574, opened 2016-03-04
Removed ViewEncapsulation.None (unnecessary)

jesperronn:
I wondered what this did and why we have it in the seed project?

ViewEncapsulation.None tells Angular not to use shadow DOM

Does  this make a difference? Did it slow down the code/tests/runtime?

My suggestion is to remove it.

More info http://blog.thoughtram.io/angular/2015/06/29/shadow-dom-strategies-in-angular2.html


mgechev:
Yes, lets simplify the example.

PR #595, opened 2016-03-08
Replace hardcoded tools path by TOOLS_DIR constant

cm0s:
I needed to change the TOOLS_DIR value and discovered it was still hardcoded in some files. Just corrected that.


mgechev:
Thanks for the PR!

PR #597, opened 2016-03-08
(build.index.prod) Fixed broken injection under windows

Nightapes:
Problem: 
When running 'run build.prod' under windows, the injected URL's have backslashes instead of slashes. 
The dev build is not effected. 

Before fix:
```html
 <!-- inject:css -->
 <link rel="stylesheet" href="css\all.css?1457433908204">
 <!-- endinject -->
```

After fix:
```html
 <!-- inject:css -->
 <link rel="stylesheet" href="css/all.css?1457433908204">
 <!-- endinject -->
```

I added 'slash' to the build.index.prod.ts and also re-factored it to be similar to build.index.dev.ts


mgechev:
Thanks for the catch!

PR #601, opened 2016-03-08
Small change according to airbnb guide

tarlepp:
https://github.com/airbnb/javascript#iterators-and-generators


mgechev:
Why not `forEach` instead, since it implies side-effect?

tarlepp:
Both of those will work in this case. Basically just matter of opinion. If you prefer .forEach I'll just quick change to this PR.

mgechev:
Yes, both will work as well as the `for-loop`. I agree with you that using high-order functions looks more elegant but we can take it even further and make it better from semantical point of view since `forEach` implies a side effect, where `map` is considered more like a pure function because of the new collection it returns.

PR #603, opened 2016-03-08
Minor change according to Airbnb JavaScript Style Guide

tarlepp:
https://github.com/airbnb/javascript#whitespace--around-keywords


ludohenin:
Can't we set that one for tslint as well ?

tarlepp:
That is very good idea, airbnb providers linter configs that we could use https://github.com/airbnb/javascript/tree/master/linters

PR #633, opened 2016-03-15
docs(readme): note about advanced seed option

NathanWalker:
@mgechev My thinking here is many could learn how to extend this seed from examining the advanced seed and/or if they are seeking a more advanced seed option, this would give them a path to find it easier. The advanced seed is really a nice compliment to this seed :+1:  It's not for everybody but definitely an option that should go hand-in-hand with this seed.


d3viant0ne:
And soon to be "advanced seed + Material2" muhuhuhuh

Issue #579, opened 2016-03-04
chore(tools): Improve deps, assets and system in project.config


ludohenin:
Extensibility of the config members is too complex ...

- [ ] DEV_NPM_DEPENDENCIES / PROD_NPM_DEPENDENCIES
- [ ] SYSTEM_CONFIG_DEV & SYSTEM_BUILDER_CONFIG
- [ ] APP_ASSETS

... I'll work out something to make it more elegant than this:

```ts
export class ProjectConfig extends SeedConfig {
  //...

  DEV_NPM_DEPENDENCIES: InjectableDependency[] = [
    ...this.DEV_NPM_DEPENDENCIES,
    ...this.normalizeDependencies([
      { src: 'materialize-css/dist/css/materialize.min.css', inject: true }
    ])
  ];

  PROD_NPM_DEPENDENCIES: InjectableDependency[] = [
    ...this.PROD_NPM_DEPENDENCIES,
    ...this.normalizeDependencies([
      { src: 'materialize-css/dist/css/materialize.min.css', inject: true }
    ])
  ];

  APP_ASSETS: InjectableDependency[] = [
    { src: `${this.ASSETS_SRC}/prism.min.js`,  inject: 'libs' },
    { src: `${this.ASSETS_SRC}/prism.css`, inject: true },
    ...this.APP_ASSETS
  ];

  DEV_DEPENDENCIES = this.DEV_NPM_DEPENDENCIES.concat(this.APP_ASSETS);
  PROD_DEPENDENCIES = this.PROD_NPM_DEPENDENCIES.concat(this.APP_ASSETS);

  FONTS_SRC = [
    'node_modules/materialize-css/dist/font/**/*'
  ];

  SYSTEM_CONFIG_DEV = {
  defaultJSExtensions: true,
    paths: {
      [this.BOOTSTRAP_MODULE]: `${this.APP_BASE}${this.BOOTSTRAP_MODULE}`,
      'config': `${this.APP_ROOT}config`,
      'angular2/*': `${this.APP_BASE}angular2/*`,
      'rxjs/*': `${this.APP_BASE}rxjs/*`,
      '*': `${this.APP_BASE}node_modules/*`
    },
    packages: {
      angular2: { defaultExtension: false },
      rxjs: { defaultExtension: false },
      moment: { main: 'moment.js' },
      lodash: { main: 'lodash.js' }
    }
  };

  SYSTEM_CONFIG = this.SYSTEM_CONFIG_DEV;

[...]

PR #644, opened 2016-03-17
chore: simplify config, deprecates npm_deps

### LINKED ISSUE #629: Merge DEV_NPM_DEPENDENCIES and PROD_NPM_DEPENDENCIES
It is confusing and redundant to have two different arrays of injectable dependencies.

The biggest current benefits of these two configuration options are:

- Faster loading of the development build (we're able to inject bundles instead of load the individual dependencies one by one, what the prod build does).
- We're able to use pre-minified version of some of the dependencies for the production build. This can eventually save some bandwidth if the files are minified in a special way (using Closure Compiler or uglifyjs wth mangling).

Generally only a single array of dependencies will be enough, and the difference in the treatment of the files could come from the configuration of the SystemJS builder that we use.

For some advanced builds it could be better to preserve both configuration options. My suggestion is to keep `PROD_DEPENDENCIES` with value `null` by default. If the user of the seed needs more custom behavior she can define `PROD_DEPENDENCIES` array and get the current behavior.

// cc @ludohenin @d3viant0ne

mgechev:
Fix #629

Still haven't implemented the setters since they will introduce breaking changes and we need to discuss them. Basically, each individual `InjectableDependency` already includes `env` and `vendor` properties. By default all each dependency belongs to both environments (which could be more that is why is used a dependency array instead of `dev`, `prod` and `both`). Once we intent to get the dependencies using `DEPENDENCIES` we resolve all of them which belong to `npm_modules` and merge the result with the ones from `APP_ASSETS` that belong to the environment.

@ludohenin @d3viant0ne let me know if you find anything wrong.

Also, lets discuss the setters for keeping backwards compatibility.


tools/config/project.config.ts
```diff
@@ -7,21 +8,19 @@ export class ProjectConfig extends SeedConfig {
...
     // this.APP_TITLE = 'Put name of your app here';
-    let additional_deps: Array<any> = [
+    let additional_deps: InjectableDependency[] = [
       // {src: 'jquery/dist/jquery.min.js', inject: 'libs'},
       // {src: 'lodash/lodash.min.js', inject: 'libs'},
     ];
 
-    this.DEV_NPM_DEPENDENCIES = this.DEV_DEPENDENCIES.concat(normalizeDependencies(additional_deps));
-    this.PROD_NPM_DEPENDENCIES = this.PROD_NPM_DEPENDENCIES.concat(normalizeDependencies(additional_deps));
+    const oldDeps = this.NPM_DEPENDENCIES;
```
d3viant0ne: I get what we are doing here but is there a better word than "oldDeps" to convey intent? I can see the future issue questions already.

d3viant0ne:
Outside of me knit picking naming conventions, I think it's a solid change. 

I have seen a few people modify the structure of the project.config (another concat iteration) to pull in other technologies. Something to keep in mind, this is a semver Minor but in reality it's a semver Major for a few projects I can think of off the top of my head. We should add a release note of some sort.

ludohenin:
I don't think it completely fixes #579 but looks good for dependencies.

tools/config/project.config.ts
mgechev: Yes, `seedDeps` sounds much better.

mgechev:
@ludohenin yes, it only partially solves it. It mostly deals with #629

tools/config/project.config.ts
mgechev: I'll update it tomorrow and assign the issue to you.

ludohenin:
I've mentioned it once already but I find it cleaner to remove the `constructor` method (it's useless in our case) and to have `project.config` like this. Don't you think so ?
```ts
export class ProjectConfig extends SeedConfig {
  PROJECT_TASKS_DIR = join(process.cwd(), this.TOOLS_DIR, 'tasks', 'project');

  // APP_TITLE = 'Put name of your app here';
  additional_deps: InjectableDependency[] = [
    // {src: 'jquery/dist/jquery.min.js', inject: 'libs'},
    // {src: 'lodash/lodash.min.js', inject: 'libs'},
  ];

  oldDeps:InjectableDependency[] = this.NPM_DEPENDENCIES;

  NPM_DEPENDENCIES = this.oldDeps.concat(this.additional_deps);

  APP_ASSETS: InjectableDependency[] = [
    // {src: `${this.ASSETS_SRC}/css/toastr.min.css`, inject: true},
    // {src: `${this.APP_DEST}/assets/scss/global.css`, inject: true},
    { src: `${this.ASSETS_SRC}/main.css`, inject: true },
  ];
}
``

**Comment by mgechev** [comment]
@ludohenin yes, it looks cleaner. I don't like that `seedDeps` becomes an instance member.

**Comment by ludohenin** [comment]
oops, there are conflicts, would you rebase ?

**Comment by ludohenin** [comment]
@mgechev Yes you're right about instance members. Then we have to keep has it is.

**Comment by mgechev** [comment]
@ludohenin in the current PR I don't like how we treat the `PROD_NPM_DEPENDENCIES`, etc. their setters don't do anything and we are not backwards compatible.

**Comment by mgechev** [comment]
@ludohenin I think it looks better now. Let me know if you find anything wrong.

File: tools/config/seed.config.ts [hunk 2]
```diff
@@ -1,9 +1,15 @@
...
+interface Environments {
+  DEVELOPMENT: string;
+  PRODUCTION: string;
+  [key: string]: string;
+}
+
+const ENVIRONMENTS: Environments = {
   DEVELOPMENT: 'dev',
   PRODUCTION: 'prod'
 };
```
ludohenin: To make `ENVIRONMENTS ` interface "usefull", we may have to move `ENVIRONMENTS ` constant as a class instance member, I suppose. And interface should be moved to `seed.config.interfaces`.

tools/config/seed.config.ts
```diff
@@ -78,9 +74,49 @@ export class SeedConfig {
...
+    this.NPM_DEPENDENCIES = notDev.concat(value);
+  }
+  set PROD_NPM_DEPENDENCIES(value: InjectableDependency[]) {
+    console.warn('The property "PROD_NPM_DEPENDENCIES" is deprecated. Consider using "DEPENDENCIES" instead.');
+    const notProd = this.NPM_DEPENDENCIES.filter(d => !filterDependency(ENVIRONMENTS.PRODUCTION, d));
+    this.NPM_DEPENDENCIES = notProd.concat(value);
+  }
 
+  get DEPENDENCIES(): InjectableDependency[] {
+    return normalizeDependencies(this.NPM_DEPENDENCIES.filter(filterDependency.bind(null, this.ENV)))
```
ludohenin: Considering the above comment, we could just `bind` to this, the `ENVIRONMENTS` and `ENV` would be available.
Just for the sake of simplifying a bit

tools/config/seed.config.ts
ludohenin: Moving `ENVIRONMENTS ` will allow extensibility.

tools/config/seed.config.ts
ludohenin: No, forget about this comment ...

tools/tasks/seed/build.html_css.ts
```diff
@@ -4,7 +4,7 @@ import * as merge from 'merge-stream';
 import * as autoprefixer from 'autoprefixer';
 import * as cssnano from 'cssnano';
 import {join} from 'path';
-import {APP_SRC, TMP_DIR, CSS_PROD_BUNDLE, CSS_DEST, APP_DEST, BROWSER_LIST, ENV, getEnvDependencies} from '../../config';
```
ludohenin: Maybe a deprecation warning is missing on `getEnvDependencies` method of `SeedConfig` class.

tools/config/seed.config.ts
```diff
@@ -50,26 +56,16 @@ export class SeedConfig {
...
     { src: 'es6-shim/es6-shim.js', inject: 'shims' },
     { src: 'systemjs/dist/system.src.js', inject: 'shims' },
     { src: 'angular2/bundles/angular2-polyfills.js', inject: 'shims' },
-    { src: 'rxjs/bundles/Rx.js', inject: 'libs' },
-    { src: 'angular2/bundles/angular2.js', inject: 'libs' },
-    { src: 'angular2/bundles/router.js', inject: 'libs' },
-    { src: 'angular2/bundles/http.js', inject: 'libs' }
-  ]);
-
-  PROD_NPM_DEPENDENCIES: InjectableDependency[] = normalizeDependencies([
```
ludohenin: Normalizing here is redundant with the normalization done with `get DEPENDENCIES()`

ludohenin:
@mgechev have added some comments. Otherwise LGTM.

Issue #648, opened 2016-03-18
chore(update): Update to Beta 10


d3viant0ne:
Looks like 2.0.0-beta.10 has a few issues to work out

```
angular2/src/common/directives/ng_plural.d.ts(2,10): error TS2305: Module angular2/src/common/directives/ng_switch"' has no exported member 'SwitchView'.
angular2/src/core/zone/ng_zone_impl.d.ts(9,44): error TS2304: Cannot find name 'ZoneSpec'.
angular2/src/core/zone/ng_zone_impl.d.ts(30,28): error TS2304: Cannot find name 'ZoneDelegate'.
angular2/src/core/zone/ng_zone_impl.d.ts(30,77): error TS2304: Cannot find name 'Task'.
angular2/src/core/zone/ng_zone_impl.d.ts(31,24): error TS2304: Cannot find name 'ZoneDelegate'.
angular2/src/core/zone/ng_zone_impl.d.ts(32,25): error TS2304: Cannot find name 'ZoneDelegate'.
angular2/src/core/zone/ng_zone_impl.d.ts(32,82): error TS2304: Cannot find name 'HasTaskState'.
angular2/src/core/zone/ng_zone_impl.d.ts(33,29): error TS2304: Cannot find name 'ZoneDelegate'.
```

jesperronn:
@d3viant0ne this seems related to angular itself, so it would probably speed things up to add an issue to the Angular issue tracker

jesperronn:
@d3viant0ne This looks to be the error reported in https://github.com/angular/angular/issues/7656.

mgechev:
Yes, here it is again https://github.com/angular/angular/issues/7660

d3viant0ne:
Yeah, I went and looked and at the time there was at least one related issue.

It's open along with the PR so we don't get 3 more PR's trying to update to beta.10

mgechev:
Today I'll merge your beta.10 PR, tag and update to beta.11.

PR #649, opened 2016-03-18
chore(deps): Update to Angular2 Beta 10

d3viant0ne:
- Update to 2.0.0-beta.10
  - Adds es6-promise
  - Update to zonejs 0.6.4

> There are a few issues that need to be resolved before this lands. I'll watch upstream and update the branch when it's ready. (See the comment in #648)


antonybudianto:
beta 11 is released

PR #667, opened 2016-03-22
chore(deps): Update to rxjs 5.0.0-beta.3

### LINKED ISSUE #665: chore(deps): update to rxjs 5 beta.3
Per the discussion in #664, Beta.2 is missing observable.of which would be moderately annoying if you actually wanted to use the Rx package.

Bumping the Rx version to beta 3 doesn't appear to have any Angular2 side effects outside of much faster build times and less memory consumption

d3viant0ne:
- Performance improvments in watch cycle
  - Performance improvments in memory consumption
  - Bundle in beta.2 is missing Obeservable.of

resolves #665

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "es6-promise": "^3.1.2",
     "es6-shim": "0.33.3",
     "reflect-metadata": "0.1.2",
-    "rxjs": "5.0.0-beta.2",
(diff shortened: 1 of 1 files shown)


mgechev:
It is not compatible with beta.9, which resides in the current master.

Issue #688, opened 2016-03-25
Add --color to Gulp tasks?


brian428:
I'm not sure if this would cause any problems, but I was wondering if you'd consider adding the --color flag to the Gulp tasks called from the npm scripts?

E.g. instead of this:
"build.dev.watch": "gulp build.dev.watch",

It would be:
"build.dev.watch": "gulp build.dev.watch --color",

This way, you always see the color-formatted output. In my experience, without this, you don't see colors in the console output when running the command from within something like IntelliJ IDEA.

Obviously, this is minor. But if you don't think it would cause problems, it might be nice to include?

mgechev:
There will be no collisions with existing flags. Do you want to open a PR with that change?

brian428:
Sure, I'll try to do that today.

mgechev:
@brian428 I added a wiki [page here](https://github.com/mgechev/angular2-seed/wiki/FAQ). Feel free to modify it if it doesn't sound good to you.

PR #690, opened 2016-03-25
#688 add --color flag to npm script calls to Gulp

brian428:


mgechev:
AppVeyor failed on any reason (doesn't seems related to your changes).

natarajanmca11:
I am getting the same error in my machine (MacBookPro with Windows 10 OS). But in OSX, its fine.
Error Message :  Not compatible with your operating system or architecture: fsevents@1.0.9

mgechev:
@brian428 it seems the flag breaks something on Windows :disappointed: 

We can add a FAQ section in the wiki which explains how you can enable coloring with the `--color` flag. Would you add such?

brian428:
OK, I'll add this to the wiki instead. It's just odd, because I'm on Windows and this works fine. Strange.

natarajanmca11:
This issue is not because of `--color`.  
Please refer the issue https://github.com/npm/npm/issues/9204

PR #703, opened 2016-03-31
chore: angular beta.13, rxjs beta.4, node 5+, npm 3+

NathanWalker:
This introduces a debatable change for sure.

My arguments for changing to support only node 5+/npm 3+:

* Avoids peer dependency warnings which cause build failures on travis.
* Simplicity by supporting only node 5.1 on travis.
* Most new users will likely be using node 5+/npm 3+ combo.
* Allows for more progressive updates to occur more frequently because we wouldn't be held back by silly peer dependency warnings that 99% have no effect on successful build.

By maintaining support for < node 5, the seed will continue to be hindered by things like this:
```
npm WARN angular2@2.0.0-beta.13 requires a peer of rxjs@5.0.0-beta.2 but none was installed.
```

This is just a *warning* and not an error. Everything works well and afaik, rxjs beta.4 is better in many regards and continues to get better everyday.

/cc @mgechev @ludohenin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: node_js
 node_js:
-- '4.0'
-- '4.1'
 - '5.1'
 sudo: false
 services:
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ It is something similar to the Angular Quick Start but does the entire build wit
 
 # How to start
 
-**Note** that this seed project requires node v4.x.x or higher and npm 2.14.7.
+**Note** that this seed project requires node v5.x.x or higher and npm 3 or higher.
 
 In order to start the seed use:
 
--- a/package.json
+++ b/package.json
@@ -105,12 +105,12 @@
     "yargs": "^4.2.0"
   },
   "dependencies": {
-    "angular2": "2.0.0-beta.12",
+    "angular2": "2.0.0-beta.13",
     "es6-module-loader": "^0.17.8",
     "es6-promise": "^3.1.2",
     "es6-shim": "0.35.0",
     "reflect-metadata": "0.1.2",
-    "rxjs": "5.0.0-beta.2",
+    "rxjs": "5.0.0-beta.4",
     "systemjs": "~0.19.18",
     "zone.js": "^0.6.6"
   }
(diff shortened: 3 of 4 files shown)


asgarddesigns:
It's a good question, my 2c is: 

Given that 5.x branch is regarded stable (albeit not the current LTS), it seems reasonable to expect that people using a bleeding edge beta, are willing to use _at least_ the latest stable primary tools.

Also, since there is plenty of tools to manage multiple versions, is there any valid reason someone wouldn't be able to use latest tools?

mgechev:
@NathanWalker your suggestion makes sense, sticking to rxjs beta.2 drags us a bit backwards because of the issues in this release and the memory leaks that it causes.

On the other hand, I am not sure how dropping support for the current node LTS will influence on developers using the seed for corporate projects. Lets keep the discussion for a few more days.

Bigous:
The problem with the peer occurs if you don't put all the peer dependencies on your project, right? If you put them there, no problem occurs, correct?

Bigous:
In this case, you are updating your rxjs to beta.4, but angular beta.13 depends on rxjs beta.2... I don't think it's a good idea not following the angular dependencies.

About switching the node version to 5+ I agree. We use it just to build and test in development, not in production... so, why not?

ludohenin:
@NathanWalker Is there an issue at supporting node 4+ but recommending to use 5+ ?
I'm a bit on @mgechev side here, corporation ave rules ... you know :)

d3viant0ne:
@NathanWalker - We need to support the LTS which is at the moment 4.x. While I use 5.9 locally, I'd never run it in production which means I wouldn't be able to use the seed as a base for any of my production apps.

As far as rxjs goes, Angular2 works fine on beta.4 of rxjs. All of the test suites pass and so on.

@Bigous - rxjs beta.2 has memory issues and sucks and any project at scale, i'm running beta4 with angular2 & the seed for two large work project being built right now and there is no issue.

d3viant0ne:
And for reference, i'm not suggesting that we push the seed to rxjs beta.3/4 only that it is possible to make it all work.

The issues related to angular with rxjs > beta.2 can be found here - angular/angular#7719

So you "can" push angular2 to TS 1.8 / rxjs beta.4 but it requires a bit of hacking to make it work.

NathanWalker:
Thanks for the comments everybody, yep let's close this. The main reason was really to get to rxjs beta.4 since as @d3viant0ne mentions has memory leaks and is unfortunate angular beta.13 is locked to beta.2 at moment. Then there's the fact that ci builds fail with node < 5 when peer deps aren't met, which seems silly since they are warnings only and not failures.

Issue #712, opened 2016-04-03
prod build broken


robbatt:
**Steps to reproduce and a minimal demo of the problem**
win7 machine, after npm install, run 
`npm run build.prod`

**Current behavior**
throws error in task build.bundles.app
```
Unhandled rejection Error: ENOENT: no such file or directory, open '<path>\angular2-seed\node_modules\dist\package.json'
            at Error (native)
        Error loading <path>/angular2-seed/node_modules/dist/package.json

```

**Expected/desired behavior**
build runs through, more info where this is coming from ... seems due to a misconfiguration in seed.config.ts > SYSTEM_BUILDER_CONFIG (misleading by the way, SYSTEM_CONFIG_PROD would be more helpful)

**Other information**

PR #713, opened 2016-04-03
fix prod build #712

robbatt:
1bd5691 did not fix it, removing `packageConfigPaths` in `SYSTEM_CONFIG_DEV` did not fix it, but also did not break the dev build, so I'm not sure for what this was needed. Doing the same thing in `SYSTEM_BUILDER_CONFIG` fixed the prod build though. Thanks @mgechev for the fast response and the right direction!


mgechev:
Yeah, thanks for the catch... https://github.com/mgechev/angular2-seed/commit/1bd56914e7430a69e4b97f24096e9ae13266e8d4 actually broke it.

PR #707, opened 2016-04-01
feat(build): add support for angular/progressive

mgechev:
@jeffbcross @alxhub I'd appreciate your feedback here.

PR #714, opened 2016-04-04
Update service worker

domfarolino:
Old caches will be deleted when a new service worker gets activated, and it is easy to bump the cache version.

SW [cannot cache POST requests](http://stackoverflow.com/questions/35270702/) and no longer registers them as an error (see request method check in the `fetch` event listener).

SW will only go to network if it needs to, and is capable of working offline

*This can be a current sw solution until full progressive support is enabled


mgechev:
Angular 2 seed drops this file in https://github.com/mgechev/angular2-seed/pull/707.

We are planning to support angular/progressive which uses SW under the hood.

PR #729, opened 2016-04-06
Set PROJECT_ROOT to root of the project

ip512:
PROJECT_ROOT does not seems to be used somewhere in the seed, and it points to tools folder.
I propose to correct it to point to the root folder, if someone wants to use it to add files to project's root with a gulp task.


mgechev:
There's a merge conflict. Would you rebase and squash?

ip512:
I removed the normalize function, it should be fine now

PR #737, opened 2016-04-07
fix(build.bundles): add packageConfigPaths so bundlers will work with modules using package.json

njs50:
The changes/reverts from #712 left this broken for prod builds again. 

From the looks of the error messages the underlying issue was related to the paths being broken when they were normalized on windows, I think it may of been dropping the * which is supposed to be replaced by the npm module name.

I don't have a windows dev machine to test this on, but this seems to work okay on my mac. I like to imagine it will fix the windows issue from #712 too since it should be providing the full path in the correct format for the OS.


mgechev:
Tested it on Windows, it seems it works fine.

PR #750, opened 2016-04-12
refactor(app): update directory structure

### LINKED ISSUE #749: app_base causing issues when serving content statically from a subfolder
The current setup works great if you serve the prod content one of the ways:

- Using node with the serve.prod command.

- Running it on a web server statically but the domain has to be at the root level.

However, if you want to serve it by a web server through a subdomain, like so: www.example.com/content/, it will not load the style and js assets from index.html.  This is because it uses the app_base to serve this content which the seed wants you to setup before packaging for prod.  But most of the time the app could be deployed in a bunch of different testing and live locations so there is no way to preset that app base.

So the real question is what is the benefit of having the app base if the app can just load it all by using relative paths?

Here is the solution that I did until there is something better:
**index.html**
Removed following lines:
`<base href="<%= APP_BASE %>">`

**tools/tasks/seed/build.index.prod.ts**
Removed APP_BASE by changing this line:
From this:
`arguments[0] = APP_BASE + path.slice(3, path.length).join(sep) + '?${Date.now()}';`
To this:
`arguments[0] = path.slice(3, path.length).join(sep) + '?${Date.now()}';`

mgechev:
## Introduction

The proposed directory structure is as follows:

```
.
├── LICENSE
├── README.md
├── appveyor.yml
├── gulpfile.ts
├── karma.conf.js
├── package.json
├── protractor.conf.js
├── src
│   └── client
│       ├── app
│       │   ├── +about
│       │   │   ├── components
│       │   │   └── index.ts
│       │   ├── +home
│       │   │   ├── components
│       │   │   └── index.ts
│       │   ├── components
│       │   └── shared
│       │       ├── index.ts
│       │       └── services
│       │           ├── name-list.service.spec.ts
│       │           └── name-list.service.ts
│       ├── assets
│       │   └── svg
│       │       └── more.svg
│       ├── css
│       │   └── main.css
│       ├── hot_loader_main.ts
│       ├── index.html
│       ├── main.ts
│       ├── sw.js
│       └── tsconfig.json
├── test-main.js
└── tools
```

## Properties

- With the new directory structure, inside `src` we can place an application server, NativeScript app, ReactNative app, etc.
- CSS and assets are in different directories. This keeps the paths between development and production.
- The directories which can be considered lazy-loading boundaries (bounded contexts) are prefixed with `+`, which brings extra semantics.
- Introduces a facades with `index.ts` files inside of each bounded context as well as in `shared`.

Resolves #749

I'd love feedback.

// cc @ludohenin @d3viant0ne


d3viant0ne:
//cc @NathanWalker 

I'm assuming the additional level of nesting is to facilitate a "server"
Are we intending on supporting the server builds via the "project" section of the tools?

If so, it will probably be a good idea to think about and document an high level answer as the question issues will come.

Also, this may be the right time to add a dev-upstream branch and start targeting that for PRs. It will help extended fork maintainers get ready for changes coming in a pending release. I purposefully picked an name !== develop to avoid conflicts with extended repos already using some sort of "git flow"

d3viant0ne:
@mgechev - I'm all about using bounded context. I'm wondering if " + " is the best way to convey that idea or is this a standard convention and I simply didn't know.

I read your blog post before looking at the associated repo. Looking at the repo, the first thing that popped into my head was, "wtf is the + all about" when looking at the directory structure.

NathanWalker:
I really like this. It would make it possible to bring the {N} app directory here:
https://github.com/NathanWalker/angular2-seed-advanced/tree/master/nativescript
Inside `src` and make it easier to incorporate into the seed build.

Worth mentioning that {N} has it's own build system with TypeScript (https://github.com/NativeScript/nativescript-dev-typescript) and maybe there's some interesting integration points we can build into the `tools` to work it into the seed build as well with this restructure in mind.

I also really like the `+folder` convention to denote lazy-loading boundaries. I haven't seen it before, but I like it. 

I never had an issue with `css` being inside `assets` and actually have a large private project fork where I'm using `.LESS` inside `assets` and it works well.

Lastly maybe it's just personal preference, but I've never seen much value to having:
`about/components`, `home/components`, etc.
+
`components`
For simplicity, it would seem to be cleaner to just have the sample components under `components` and not scattered throughout `src/client/app`. I've taken this approach here:
https://github.com/NathanWalker/angular2-seed-advanced/tree/master/src
Top level `components` (nested in the top level `components` folder there) to me are essentially the `primary` pages/components of the app. Those `primary` pages/components may be made up of dozens to hundreds of other components pulled from (in the advanced seed case) `frameworks`, however here they could come from `shared` or just in the nested `feature` folders `about`, `home`, etc.

To me, just more clear. Allows one to go to a single folder `components` to deal with the primary (master level if you will) components that are wired usually in with the router (lazy loaded or not). Then those reference their pieces from the infinitely scalable `shared`, `frameworks` (advanced seed), or any other folder that makes sense to grow the component tree in.

ludohenin:
I agree with @NathanWalker about the scattered component structure.
Is it such a good idea to tight naming to implementation/configuration for lazy-loaded context ? But I like the naming though

nareshbhatia:
What if we renamed the `css` folder to `styles`? This allows us to keep styles written in css, scss, less or whatever else in this folder. Granted that angular-seed focuses on css only, but other extensions like angular2-seed-sass now have a good home for sass files. If yes, we should also change variables like `CSS_SRC` to `STYLES_SRC`.

d3viant0ne:
@nareshbhatia - Solid idea. In reality, the list of people that are going to build with generic css is going to be small. This rather benign change will save everyone else a bunch of refactoring / merge conflict headaches when pulling from upstream.

nareshbhatia:
While on this topic I have a question about this code in `seed.config.ts`:

    // Declare local files that needs to be injected
    APP_ASSETS: InjectableDependency[] = [
      { src: `${this.CSS_SRC}/main.css`, inject: true, vendor: false }
    ];

Is `APP_ASSETS` meant for injecting CSS only? If yes, we should change the name to `APP_STYLES`.

In addition, we should probably document `InjectableDependency` interface properly. It can be pretty confusing. For example, what does vendor = true mean? Does it mean that we are automatically going to look for it in node_modules? Similarly what does inject = true mean vs. some string?

ludohenin:
@nareshbhatia 
> Is APP_ASSETS meant for injecting CSS only?

No, this is also for JS files to be loaded using the `script` tag.

> In addition, we should probably document InjectableDependency interface properly [...]

This is slightly out-of-date but there is something here: https://github.com/mgechev/angular2-seed/wiki/Add-external-scripts-and-styles

nareshbhatia:
@ludohenin, thanks for the reference. It clears bunch of my questions. It may be useful to write this doc from the perspective of the interface rather than the use case of external scripts and styles. I think the former will allow us to document every possibility exhaustively. Just a thought.

PR #768, opened 2016-04-15
fix home component paths

alexweber:
update directory prefix for template & style urls for consistency with: https://github.com/mgechev/angular2-seed/blob/master/src/client/app/%2Babout/components/about.component.ts

### DIFF
--- a/src/client/app/+home/components/home.component.ts
+++ b/src/client/app/+home/components/home.component.ts
@@ -5,8 +5,8 @@ import {NameListService} from '../../shared/index';
 
 @Component({
   selector: 'sd-home',
-  templateUrl: 'app/+home/components/home.component.html',
-  styleUrls: ['app/+home/components/home.component.css'],
+  templateUrl: './app/+home/components/home.component.html',
+  styleUrls: ['./app/+home/components/home.component.css'],
   directives: [FORM_DIRECTIVES, CORE_DIRECTIVES]
 })
 export class HomeComponent {


alexweber:
or maybe it's the other way round? :)

TheDonDope:
To `./` prefix or not to prefix the directory confuses me too, to be honest. 
In https://github.com/mgechev/angular2-seed/blob/master/tools/config/seed.config.ts `APP_SRC` points to `src/client`. So if this is the "root" directory for the sources my guess would be that the path from there on, would be written as `./app/+home/home.component.css`. Still it feels confusing to write it that way since `.` denotes the current directory of the file (in this case `home.component.ts`), doesn't it? :confused: 

Can someone please clarify? Thank you very much :)

d3viant0ne:
@TheDonDope - You have to think about where that is actually being processed from which changes depending on dev / prod / test which I would think is why it is fully qualified from any of the following..

```
  DIST_DIR             = 'dist';
  DEV_DEST             = `${this.DIST_DIR}/dev`;
  PROD_DEST            = `${this.DIST_DIR}/prod`;
  TMP_DIR              = `${this.DIST_DIR}/tmp`;
  APP_DEST             = `${this.DIST_DIR}/${this.ENV}`;
```

mgechev:
@TheDonDope yes, without prefix it may make more sense. I'll refactor all the files today.

TheDonDope:
@mgechev, @d3viant0ne Thank you guys for the info & clear up :)

PR #772, opened 2016-04-17
Will clean out the dist/dev directory contents

benpriebe:
The karma.conf.js has a file watcher on:

```
{ pattern: 'dist/dev/**/*.js', included: false, watched: true }
```

When working with the build.test.watch gulp task, every file change runs the clean.dev task which currently deletes the dist/dev directory. The new build is then dropped in to a new /dist/dev directory which the karma test file watcher does not know about.

This fix keeps the /dist/dev directory but cleans out its contents.


d3viant0ne:
And also break the travis & appveyor builds

Issue #722, opened 2016-04-06
"src" in "exclude" in tsconfig.json disables TypeScript settings in Sublime editor


htdt:
Sublime Text editor (build 3103) with [Typescript plugin (0.1.12)](https://github.com/Microsoft/TypeScript-Sublime-Plugin) doesn't apply `tsconfig.json` settings for `.ts` files from `src` folder.
Due to this, the editor highlights wrong errors. For example,
> Experimental support for decorators is a feature that is subject to change in a future release. Specify '--experimentalDecorators' to remove this warning.

If `src` is removed from `exclude` in `tsconfig.json`, the editor is working as intended.

mgechev:
This seems like a limitation in Sublime Text's TypeScript plugin and not related to the seed.

As temporal workaround you can remove `src` from `exclude` but generally this config is required for the speed-up workaround for Windows.

zachallen:
It appears that the secondary tsconfig.json file located in src/client messes with Sublime. If you open up one of the e2e scripts (about.component.e2e.ts for example) it doesn't understand what `describe`, `browser`, `element`, etc are. If you delete that tsconfig.json file, it does.

d3viant0ne:
@zachallen - In my experience the support from best to worst === VisualStudio Code / Atom / WebStorm / Vim / Sublime for TypeScript

The support in sublime is lacking to say the least.

htdt:
The editor (tested in Sublime and VS Code) cannot get typings with `tsconfig.json` in `src` (`src\client`) folder. To solve it, create `typings.d.ts` with `/// <reference path="../typings/browser.d.ts" />` in `src` (`src\client`).

Sjiep:
I can confirm that the issue mentioned above by @htdt also is the case in Intellij (Webstorm) and adding the `typings.d.ts` into `src/client` resolves the issue. Any reason why this file should not be part of the seed?

PR #776, opened 2016-04-18
Fixes component spec tests

TheDonDope:
Removes the `expect` import from the spec files, since those import
NgMatchers which only contain Angular specific jasmine matchers.
However the current tests just use standard jasmine matchers.

---

Currently, the `expect` import leads to importing `NgMatchers` (https://angular.io/docs/ts/latest/api/testing/NgMatchers-interface.html), which defines the following Angular specific jasmine matchers:

- `not`
- `toBeAnInstanceOf`
- `toBePromise`
- `toContainError`
- `toHaveCssClass`
- `toHaveCssStyle`
- `toHaveText`
- `toImplement`
- `toMatchPattern`
- `toThrowErrorWith`

Some of those are not yet documented in the API documentation (see above link).

Since those matchers are Angular specific and do not cover all of the currently used jasmine matchers (which makes sense, since they only provide angular specifics) i propose to remove the import and work with the default jasmine matchers.

Greetings

Dope

### DIFF
--- a/src/client/app/+about/components/about.component.spec.ts
+++ b/src/client/app/+about/components/about.component.spec.ts
@@ -1,7 +1,6 @@
 import {
   TestComponentBuilder,
   describe,
-  expect,
   injectAsync,
   it
 } from 'angular2/testing';
--- a/src/client/app/+home/components/home.component.spec.ts
+++ b/src/client/app/+home/components/home.component.spec.ts
@@ -1,7 +1,6 @@
 import {
   TestComponentBuilder,
   describe,
-  expect,
   injectAsync,
   it
 } from 'angular2/testing';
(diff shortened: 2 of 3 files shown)


htdt:
NgMatchers interface extends jasmine.Matchers, not replacing it.

TheDonDope:
@htdt Thank you for the hint! 
In that case this is an issue of my Editor. I only stumbled across this, since SublimeText (plus tslint) annotated the jasmine Matchers and threw an error like `NgMatchers does not contain a property toEqual`.

This pull request can be abandoned then.

Greetings

Dope

htdt:
also got this issue, was solved [this way](https://github.com/mgechev/angular2-seed/issues/722#issuecomment-210629563)

PR #777, opened 2016-04-18
chore(tools): expose browserSync configuration

cadriel:
Exposes the browserSync configuration to the seed configuration. Can now easily be overridden.

### DIFF
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -112,6 +112,7 @@ export class SeedConfig {
       .concat(this.APP_ASSETS.filter(filterDependency.bind(null, this.ENV)));
   }
 
+
   // ----------------
   // SystemsJS Configuration.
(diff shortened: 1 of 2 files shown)


njs50:
also fixes a bug where --base wasn't working if you do a `npm cache clean && npm install`/

For whatever reason browserSync routes with a trailing slash are currently non functional.

d3viant0ne:
@njs50  Out of sheer curiosity, what are you doing with browserSync that you need the override?

cadriel:
In our case, we want to change the default `ghostMode` option.

njs50:
it can be a bit confusing when doing stuff in one tab does the same thing in all your other tabs (or iframes). The default ghostMode relays all events between all open browsers.

njs50:
@d3viant0ne in particular our pattern library which uses our ui components, and also has examples of them running in iframes can cause weird things with it enabled.

Issue #145, opened 2015-10-12
Object doesn't support property or method "keys" IE 11


punov:
Browser IE 11,
Object is not supported property or method 'keys',
angular2.dev.js, line: 4027, row: 5
if ((new exports.Map()).keys().next)  ....

What we can do here, any possible workarounds?
Thank you

mgechev:
Its weird that https://github.com/mgechev/angular2-seed/pull/143 didn't fixes that issue.

Are you using any tag or the latest version, which is in master?

punov:
Update with new changes helped as a charm, thank you!
Sorry for bothering

navcoder:
Hi,
Has anyone checked this issue with the production build? I am facing this issue even with the seed project where npm-start (dev) version works fine on IE 11, but when I use "npm run build.prod" to create production build, it again throws the same error "Object doesn't support property or method "keys" on IE 11.
Please help its urgent.

JohnCashmore:
I have this issue as well

mgechev:
@navcoder most likely you'll need to include [this polyfill](https://github.com/mgechev/angular2-seed/blob/master/test-main.js#L1-L10) somewhere in the prod build:

JohnCashmore:
should this not be included as part of the seed build? or is it down to something @navcoder put in their  project?

mgechev:
Actually it is included in [dev and prod by default](https://github.com/mgechev/angular2-seed/blob/master/src/client/index.html#L21-L32).

Do you have this shim missing in your index.html, prod output?

JohnCashmore:
I have the same issue and it is in the prod index.html and it still happens.

On Wed, Apr 20, 2016 at 2:03 PM, Minko Gechev <notifications@github.com>
wrote:

> Actually it is included in dev and prod by default
> <https://github.com/mgechev/angular2-seed/blob/master/src/client/index.html#L21-L32>
> .
>
> Do you have this shim missing in your index.html, prod output?
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/mgechev/angular2-seed/issues/145#issuecomment-212414716>
>



-- 
Whoooop Ltd is a company registered in England and Wales. Registered
number: 8877885. VAT Registered: GB182051824. Registered office: 7
Kingsfield Crescent, Biddulph, ST8 6DW

navcoder:
[...]

PR #786, opened 2016-04-21
Including shims in the production build.

nulldev07:
This one is for the bug where the production build breaks on IE 11 because the shims are not included in the production build.

### DIFF
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -50,10 +50,10 @@ export class SeedConfig {
   CODELYZER_RULES      = customRules();
 
   NPM_DEPENDENCIES: InjectableDependency[] = [
-    { src: 'systemjs/dist/system-polyfills.src.js', inject: 'shims', env: ENVIRONMENTS.DEVELOPMENT },
-    { src: 'reflect-metadata/Reflect.js', inject: 'shims', env: ENVIRONMENTS.DEVELOPMENT },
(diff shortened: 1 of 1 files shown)


mgechev:
@nulldev07 would you also verify that it works without system polyfils and set the file to be used in development only (based on this comment https://github.com/mgechev/angular2-seed/issues/145#issuecomment-212525507).

nulldev07:
Ok, we actually do not need system.js in prod - that was my oversight. Here is the new pull request. I have tested the code(both dev and prod) on these browsers:
Edge@25.10586.0.0
Firefox@43.0.1
Chrome@49.0.2623.112 m
IE@11.162.10586.0

Flood:
Have you removed es6-shim from seed now?

mgechev:
@Flood, yes we use [`core-js`](https://github.com/mgechev/angular-seed/blob/master/package.json#L148) instead.

Flood:
Ok, thanks. So basically I don't need to import es6-shim in index?

`<script src="https://cdnjs.cloudflare.com/ajax/libs/es6-shim/0.35.3/es6-shim.min.js"></script>`

I am trying to get "fetch as google" in Webmaster tool to work, so our site can be indexed. At the moment it shows the loading screen only.

mgechev:
That's related to server side pre-rendering. Take a look at Angular Universal.

Flood:
Yes we will integrate Angular Universal, but Google should still be able to render Angular(seed) apps :)

PR #791, opened 2016-04-26
Update to release beta.16

JohnCashmore:
I've updated the project to release beta.16 and modified the tests, however it was a little rushed so not 100% they've been updated correctly.

### DIFF
--- a/src/client/app/+about/components/about.component.spec.ts
+++ b/src/client/app/+about/components/about.component.spec.ts
@@ -2,25 +2,29 @@ import {
   TestComponentBuilder,
   describe,
   expect,
-  injectAsync,
-  it
-} from 'angular2/testing';
+  inject,
+  it,
+  AsyncTestCompleter
+} from 'angular2/testing_internal';
 import {Component} from 'angular2/core';
 import {DOM} from 'angular2/src/platform/dom/dom_adapter';
 import {AboutComponent} from './about.component';
 
 export function main() {
   describe('About component', () => {
+
+
     it('should work',
-      injectAsync([TestComponentBuilder], (tcb: TestComponentBuilder) => {
-        return tcb.createAsync(TestComponent)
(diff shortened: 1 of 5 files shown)


mgechev:
LGTM

src/client/app/+about/components/about.component.spec.ts
```diff
@@ -2,25 +2,29 @@ import {
   TestComponentBuilder,
   describe,
   expect,
-  injectAsync,
-  it
-} from 'angular2/testing';
+  inject,
+  it,
+  AsyncTestCompleter
+} from 'angular2/testing_internal';
```
ludohenin: It should use `testing` module as `testing_internal` is used to test the framework itself.

src/client/app/+about/components/about.component.spec.ts
```diff
@@ -2,25 +2,29 @@ import {
...
 import {AboutComponent} from './about.component';
 
 export function main() {
   describe('About component', () => {
+
+
     it('should work',
-      injectAsync([TestComponentBuilder], (tcb: TestComponentBuilder) => {
-        return tcb.createAsync(TestComponent)
+      inject([TestComponentBuilder, AsyncTestCompleter], (tcb: TestComponentBuilder, async: AsyncTestCompleter) => {
```
ludohenin: Why is `AsyncTestCompleter` required, as per the doc it should work without it ?

ludohenin: It should actually be 
```
async(inject([TestComponentBuilder], (tcb: TestComponentBuilder) => {
  //...
}));
```

JohnCashmore: I thought this too, but it fails with 

```
TypeError: undefined is not a constructor (evaluating 'new AsyncTestZoneSpec(finishCallback, failCallback, testName)') (line 98)
```

So i refered to the source to see how its implemented in the angular source

JohnCashmore:
I've made some amends as per comments

PR #795, opened 2016-04-28
fix(tests): add fakeasync dependency to karma

JohnCashmore:
As per change log for angular 
> You will also need to add the dependency 'node_modules/zone.js/dist/fake-async-test.js' as a served file in your Karma or other test configuration.


JohnCashmore:
Sorry missed these changes yesterday, the specs are still valid, but it might be worth reviewing to see if they should be using `async(inject`

PR #799, opened 2016-04-29
chore(readme): change node requirements

JohnCashmore:
node 5 is not required


mgechev:
Yes, good point. I was doing the same in `ci-node4` branch.

Shyam-Chen:
Windows require node v5.10.1 ...

Issue #804, opened 2016-05-01
Browsersync causing a full refresh instead of injecting CSS & images. 


heavymery:
> tools/utils/seed/code_change_tools.ts
```js
let changed = (files: any) => {
  if (!(files instanceof Array)) {
    files = [files];
  }
  // if (ENABLE_HOT_LOADING) {
  //   ng2HotLoader.onChange(files);
  // } else {
  //TODO: Figure out why you can't pass a file to reload
  console.log(files);
  browserSync.reload(files);
  //}
};
```
https://github.com/mgechev/angular2-seed/commit/e3696aa15410a6befc0a37bb8732e4dc2693e924#commitcomment-16994199

@d3viant0ne 
Just as @gkalpak's [comment](https://github.com/mgechev/angular2-seed/commit/e3696aa15410a6befc0a37bb8732e4dc2693e924#commitcomment-16994199), `files.path` always be `undefined`. So, Browsersync causing a full refresh instead of injecting CSS & images. 

Because `files` will always be Array and `browserSync.reload` accept String or Array, We can call `browserSync.reload` with `files`.

>https://www.browsersync.io/docs/api/#api-reload
### Browsersync API
**.reload( arg )**
...
**arg**
Type: String | Array | Object [optional]

Then, injecting CSS & images will working correctly.

ludohenin:
This is solved by bb00b19e5fab48e00a99aecda66f881dd59a2b35

heavymery:
:metal:

PR #807, opened 2016-05-02
perf update -> css streaming

totev:
Enabling the streaming of css files handled by browsersync.
Will look through the list of changed file and if only css or scss file was changed, then all css files will be internally reloaded per stream by browsersync. If some other file type is found amongst the changed one - a standard  reload will occur.

It works nicely with scss, when a separate task compiles a concat file containing all of the styles and a normal watch will only find the scss files, as they are the ones lying in the project scope.

Note that component scoped css files will not be covered by this as they are loaded by angular directly.

Addresses #804.

### DIFF
--- a/tools/utils/seed/code_change_tools.ts
+++ b/tools/utils/seed/code_change_tools.ts
@@ -1,6 +1,6 @@
 import {BROWSER_SYNC_CONFIG} from '../../config';
 import * as browserSync from 'browser-sync';
-let path = require('path');
+import * as path from 'path';
 
 let runServer = () => {
   browserSync.init(BROWSER_SYNC_CONFIG);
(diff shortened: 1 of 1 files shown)


tools/utils/seed/code_change_tools.ts
```diff
@@ -1,5 +1,6 @@
 import {BROWSER_SYNC_CONFIG} from '../../config';
 import * as browserSync from 'browser-sync';
+let path = require('path');
```
mgechev: When type definitions are available we're using ES2015 module imports:

```ts
import {parse} from 'path';
```

or

```ts
import * as path from 'path';
```

mgechev:
Looks great! I left a small comment for consistency.

totev:
Thanks for the comment! 
Noted.

PR #813, opened 2016-05-04
is there any use?

Shyam-Chen:
CircleCI?


mgechev:
Lets keep it that way in case someone introduces CircleCI in their project.

PR #828, opened 2016-05-08
Changed the concatenation of css files to use concatCss instead of co…

gotenxds:
Currently if any files use the @import feature of css it is being ignored by the build.
This is not a problem when using files from inside the project as they are being included anyway but when using 3rd party library if it uses @import on its own files they will not be included in the prod build(as they are not present) and even if added forcefully to the project will throw runtime errors in the browser for missing files.

This pr fixes this by changing the use of gulp-concat for css files with gulp-cssConcat which handles @import statements correctly.
Also because gulp-cssConcat does not automatically minify the code I added the use of gulp-cleanCss.

### DIFF
--- a/tools/tasks/seed/build.html_css.ts
+++ b/tools/tasks/seed/build.html_css.ts
@@ -6,6 +6,7 @@ import * as cssnano from 'cssnano';
 import {join} from 'path';
 import {APP_SRC, TMP_DIR, CSS_PROD_BUNDLE, CSS_DEST, APP_DEST, BROWSER_LIST, ENV, DEPENDENCIES} from '../../config';
 const plugins = <any>gulpLoadPlugins();
+let cleanCss = require('gulp-clean-css');
 
 const processors = [
   autoprefixer({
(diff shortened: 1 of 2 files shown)


tools/tasks/seed/build.html_css.ts
```diff
@@ -6,6 +6,7 @@ import * as cssnano from 'cssnano';
 import {join} from 'path';
 import {APP_SRC, TMP_DIR, CSS_PROD_BUNDLE, CSS_DEST, APP_DEST, BROWSER_LIST, ENV, DEPENDENCIES} from '../../config';
 const plugins = <any>gulpLoadPlugins();
+let cleanCss = require('gulp-clean-css');
```
mgechev: Is `plugins.cleanCss` available?

d3viant0ne: Should be and probably a good idea to use a TS standard `import` as opposed to requiring for the sake of consistancy

mgechev: Agree. In this case in order to save work from writing manual typings `plugins.cleanCss` seems like a good idea.

gotenxds:
plugins,cssClean does not work, for some strange reason `gulpLoadPlugins` does not find it.

PR #856, opened 2016-05-11
Update directory structure in README.md

TheDonDope:
Hello guys,

i propose this commit to update the directory and file listing in the README to reflect
the current layout. I saw that before some of the file names were abbreviated with a `...`. I couldn't see a specific reason for this, so i included the full file names, as well as a listing of all the files in the directory structure. If you do not approve of this, feel free to reject the pull request :)

Greetings

Dope

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,44 +179,127 @@ Forks of this project demonstrate how to extend and integrate with other librari
 ├── package.json               <- dependencies of the project
 ├── protractor.conf.js         <- e2e tests configuration
 ├── src                        <- source code of the application
-│   ├── home
-│   │   └── components
-│   ├── index.html
-│   ├── main.ts
-│   ├── shared
-│   │   └── services
-│   │       ├── name-list...
-│   │       └── name-list...
-│   └── sw.js                  <- sample service worker
+│   └── client
+│       ├── app
+│       │   ├── +about
+│       │   │   ├── components
+│       │   │   │   ├── about.component.css
+│       │   │   │   ├── about.component.e2e-spec.ts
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -179,44 +179,127 @@ Forks of this project demonstrate how to extend and integrate with other librari
...
-│   ├── main.ts
-│   ├── shared
-│   │   └── services
-│   │       ├── name-list...
-│   │       └── name-list...
-│   └── sw.js                  <- sample service worker
+│   └── client
+│       ├── app
+│       │   ├── +about
+│       │   │   ├── components
```
mgechev: This reminds me...we have one more violation of the style guide here. Instead of grouping by type it'll be better to group by feature, i.e. we can remove the `components` directory here, same for `+home`.

For `app`, we should move `app.component.*` outside the `components` directory and move `toolbar`, `navbar`, into separate directories within a directory called `shared`. After that we need to add barrels in each lazy-loading boundary.

TheDonDope: @mgechev Shall i create a PR for that? Also another question: I know the `client` directory got introduced to enable multiple clients within the project. Yet, the seed does not (yet?) make use of that. I don't know if that'd be considered another violation to the styleguide, but i found this sample structure from the Styleguide (Style 04-06 Overall Structural Guidelines):

screenshot 2016-05-11 16 36 15

So, should we keep the `client` subfolder?

Thank you for you input!

mgechev: Lets keep it for now. I'll start a discussion with @johnpapa and share the outcome.

PR #858, opened 2016-05-11
Apply Styleguide Rules 04-* Application Structure

### LINKED ISSUE #857: Directory structure - alignment with the style guide
After a lot of discussions between we agreed upon a directory structure, which is published at https://angular.io/styleguide.

Unfortunately, angular2-seed is violating it. Here are suggestions for alignment with the standard:

> Instead of grouping by type it'll be better to group by feature, i.e. we can remove the components directory here, same for +home.

> For app, we should move app.component.* outside the components directory and move toolbar, navbar, into separate directories within a directory called shared. After that we need to add barrels in each lazy-loading boundary.

Source https://github.com/mgechev/angular2-seed/pull/856#discussion_r62858563

TheDonDope:
Applies the Styleguide rules for the application structure.
Fixes #857.

This commit is a work in progress, i could use some help, i will add some line comments and upload screenshots for the problems i currently face.

Thanks for your help!


src/client/app/app.component.ts
```diff
@@ -1,11 +1,9 @@
 import { Component } from '@angular/core';
 import { ROUTER_DIRECTIVES, Routes } from '@angular/router';
 
-import { AboutComponent } from '../+about/index';
-import { HomeComponent } from '../+home/index';
-import { NameListService } from '../shared/index';
-import { NavbarComponent } from './navbar.component';
-import { ToolbarComponent } from './toolbar.component';
+import { AboutComponent } from './+about';
```
TheDonDope: As of the Styleguide the barrel syntax (Style 04-10 Create and Import Barrels - https://angular.io/styleguide#!#application-structure )for the imports list the following example:

```
import {
  CONFIG,
  EntityService,
  ExceptionService,
  FilterTextComponent,
  InitCapsPipe,
  SpinnerService,
  ToastService
} from '../shared';
```

Within the folder `shared` there is an `index.ts` barrel file. However, if i import it like this, i get an exception at runtime:

screenshot 2016-05-11 17 49 08

With this syntax, a lookup is made for +about.js, +home.js, shared.js; yet the corresponding files are +about/index.js, +home/index.js and shared/index.js.

I dont know if that is a problem of the SystemJS Configuration.

Any help on this would be appreciated.

tools/config/seed.config.ts
```diff
@@ -21,7 +21,8 @@ export class SeedConfig {
   ENABLE_HOT_LOADING   = argv['hot-loader'];
   HOT_LOADER_PORT      = 5578;
 
-  BOOTSTRAP_MODULE     = this.ENABLE_HOT_LOADING ? 'hot_loader_main' : 'main';
+  BOOTSTRAP_DIR        = 'app';
```
TheDonDope: Here is another "issue" or "proposal" of sorts. The styleguide states
> Do put bootstrapping and platform logic for the app in a file named main.ts.

In the folder structure examples, this file resides in `src/app` (`src/app/main.ts`).
In the seed the `main.ts` this file currently resides in `src/client` (`src/client/main.ts`).
To conform to the styleguide would it be better to move this to `src/client/app`? (`src/client/app/main.ts`). This is what i went for in this first commit here.

I'm looking for input on that too :)

src/client/app/app.component.ts
mgechev: We need to change `tsconfig.json`. for now `import * as foo from '../shared/index'` is fine.

tools/config/seed.config.ts
mgechev: Yes, it should be in `src/client/app`.

TheDonDope:
Alright, i got it all working!! 🎉

mgechev:
@TheDonDope you're very welcome! Thanks for the awesome job! I'll review and merge tomorrow.

TheDonDope:
@mgechev Thank you :) I'm glad if i could help! Oh and: the appveyor build failed again :/ I recon it is the same timing/racing condition problem like yesterday. Have a nice evening!

src/client/app/+home/home.component.ts
```diff
@@ -1,12 +1,12 @@
...
 import { Component } from '@angular/core';
 
-import { NameListService } from '../../shared/index';
+import { NameListService } from '../shared/index';
 
 @Component({
   selector: 'sd-home',
-  templateUrl: 'app/+home/components/home.component.html',
-  styleUrls: ['app/+home/components/home.component.css'],
+  templateUrl: 'app/+home/home.component.html',
```
mgechev: Just a random thought - we're using commonjs which means that we can use `moduleId` with `module.id`. Not sure yet, if we should do that or keep the path absolute. `module.id` doesn't work with systemjs where we should use `__moduleName` instead.

Do you have any thoughts?

TheDonDope: Hey!
As a matter of fact
`moduleId = module.id` was in the code of the seed some while ago but got removed at some point. I'll have a look for the specific commit which removed it to see if there was any reason for this. Would the inclusion also enable the use of the barrel imports without the `/index` suffix, or would there have to be another configuration step in the `tsconfig.json` as you mentioned earlier?

ludohenin: `module.id` was removed as `cjs` specific.

TheDonDope: Hey @ludohenin, i just stumbled across this recent talk https://youtu.be/WAPQF_GA7Qg?t=5m39s by John Papa and saw that the moduleId: module.id is explicitly included within the snippet of Visual Studio Code and he furthermore explains that Angular2 makes use of this since it uses commonjs. Sorry if i got something wrong there, but doesn't that open the possiblity of re-adding module.id, considering the seed uses commonjs as well? 
Thanks for your input :)

//cc @mgechev

mgechev: Yes, our configuration is commonjs specific so we can. In case developers switch to any other module format they have to manually handle the migration (such as __moduleName for system).

PR #866, opened 2016-05-12
Document project and seed configuration

### LINKED ISSUE #861: Is difficult to understand Seed configuration file
I think is difficult to understand the seed.config.ts file, what is each thing and what is used for.

Is important because if somebody wants to add news libraries, is not clear what have to modify.

TheDonDope:
Hey guys,

as announced beforehand, here is my proposal to add inline documentation for the project.config.ts and seed.config.ts using jsdoc notation (http://usejsdoc.org/). This aims to fix #861.

Greetings

Dope

### DIFF
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -32,19 +32,19 @@ export class SeedConfig {
    * when running `npm start`.
    * @type {number}
    */
-  PORT                 = argv['port']                        || 5555;
+  PORT = argv['port'] || 5555;
 
   /**
    * The root folder of the project (up two levels from the current directory).
    */
-  PROJECT_ROOT         = join(__dirname, '../..');
+  PROJECT_ROOT = join(__dirname, '../..');
 
   /**
    * The current environment.
    * The default environment is `dev`, which can be overriden by the `--env`
    * flag when running `npm start`.
    */
-  ENV                  = getEnvironment();
+  ENV = getEnvironment();
 
   /**
    * The flag for the debug option of the application.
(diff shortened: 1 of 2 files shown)


tools/config/seed.config.ts
```diff
@@ -3,54 +3,241 @@ import { argv } from 'yargs';
...
+ */
 export class SeedConfig {
+
+  /**
+   * The port where the application will run.
+   * The default port is `5555`, which can be overriden by the  `--port` flag
+   * when running `npm start`.
+   * @type {number}
+   */
   PORT                 = argv['port']                        || 5555;
```
TheDonDope: Also the question here is if we want to keep this indentation. it looked nice when all the constants were listed among each other. It might look a little weird now, but i guess thats just a question of style preference ;D

ludohenin: Good point, it doesn't make sense to me as well now.

TheDonDope: Alright, then i'll go ahead and add another commit to change back to a normal indentation :)

TheDonDope:
Alright, the indentation is fixed and some last misspellings are corrected :) I'm done from my side.

TheDonDope:
I dont know why the appveyor build failed for the second commit, it ran fine for the first one, and the second one only removes some whitespace and alters documentation. I generally have a feeling that the appveyor build tends to fail on PRs with multiple commits (For the first one mostly succeeds but further builds tend to fail due to timing problems / racing conditions).

ludohenin:
That's Chrome webdriver that fails, IE is passing. Something with Chrome on Windows maybe ?

Issue #867, opened 2016-05-12
Add inline documentation to existing TypeScript files


TheDonDope:
Currently, there is just a little inline documentation in the existing TypeScript files (tasks, utils etc.).
Continuing the strive for more documentation it should be aimed at to add more documentation on code level.

I will sign up for this, and keep this as a tracking issue. The following sections should be documented:

- [X] Files in `src/client` (merged with https://github.com/mgechev/angular2-seed/commit/e1a32af5592beb776affc96655a5eb27f87e9867)
- [X] Files in `tools/tasks` (merged with https://github.com/mgechev/angular2-seed/commit/2b8cc50d51820b11bd72fd02b24803d7bbb967ca)
- [X] Files in `tools/utils` (merged with https://github.com/mgechev/angular2-seed/commit/32280fdba91ce3dea7d60b76528b767a65b178a0)

Greetings

Dope

P.S.: Can someone please assign this issue to me, or is an assignment only possible for write-access contributors?

d3viant0ne:
@TheDonDope - Assignment requires write access to the repository.

TheDonDope:
@d3viant0ne Ah, thanks for the clarification! Then i'll have to make do without the assignment i guess, i mostly would have wanted it as a sign of "someone is already working on that, no need to do redundant work" :)

d3viant0ne:
Can also be accomplished by a pull request 
i.e - [WIP] feat(docs): Inline Documentation

TheDonDope:
I plan on submitting separate PRs for the different tasks in this issue, to keep this issue as a "tracker" of sorts. My train of thought was:
1. PR for docs in src/client -> merge -> update task list in this issue
2. PR for docs in tools/tasks -> merge -> update task list in this issue
3. PR for docs in tools/utils -> merge -> update task list in this issue
4. Close this issue as all tasks have been finished.

TheDonDope:
:tada: 
All done :) Thanks for the reviews and merges! 
Issue can be closed now.

Greetings

Dope

mgechev:
Thanks for the great work! :-)

TheDonDope:
Glad to have been of help :)

PR #878, opened 2016-05-14
Document sources in src/client

TheDonDope:
Hey guys,

this is the PR to complete the first task of #867, providing inline documentation for the TypeScript files
within `src/client` using the jsdoc notation. 
Any feedback / change request to this PR is very much welcome :)

Greetings

Dope

### DIFF
--- a/src/client/app/+about/index.ts
+++ b/src/client/app/+about/index.ts
@@ -1,2 +1,6 @@
+/**
+ * This barrel file provides the export for the lazy loaded
+ * AboutComponent.
+ */
 export * from './about.component';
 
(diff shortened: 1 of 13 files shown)


src/client/app/+about/index.ts
```diff
@@ -1,2 +1,6 @@
+/**
+ * This barrel file provides the export for the lazy loaded
```
ludohenin: @mgechev So barrel won against facade :)

mgechev: Yeah, we had discussions about this. Barren is a special kind of a facade but with more specific semantics.

Issue #802, opened 2016-04-29
Possible Issue with inclusion of reflect-metadata in production build


brian428:
I don't think this is directly related to the seed project, but I thought I would mention it:

After this change (57d18695dfc73cd769e34947ef419237cd8b7954), I started getting some errors in my app after a production build. Traced it back to the inclusion of reflect-metadata. If I change Reflect back to only include in the dev build, things seem to work fine again.

My understanding of things like Reflect, Zone, and angular2-polyfills is limited, so I'm not exactly sure why Reflect is causing me problems (and why it's only in the production build). My app is expanded from what's in the base seed, and the error is a strange issue when working with SVGElement (something done by the Highcharts library). This makes me think it must have something to do with a shim or prototype override. But I'm not even sure if it's actually Reflect, or if it's really one of the other shims and including Reflect just causes the problem to manifest itself.

Anyway, as I said, I don't think this is really a seed project issue, but just wanted to point it out in case you wanted to mention possible issues caused by including these in production in the wiki or readme.

(That said, if you have any idea off the top of your head about what the problem might be, I'd be all ears!)

atticae:
I have the same issue in production mode and solved it by reverting the change you linked. No idea what exactly is causing the problem. I am also working with Highcharts, so it might be related to that?

brian428:
Well I doubt it is Highcharts itself, since it works fine in dev mode, and works fine in prod mode when Reflect is omitted. I'd tend to think it's something Reflect is doing, but not sure what (I know what Reflect is in general terms, but not really how it does what it does). 

If other people are seeing this, that makes me think maybe I should add an issue to the Reflect tracker...

sprell:
I am experiencing a similar issue. Seems to be due to the fact that reflect.js has "use strict" at the top which impacts all the other JavaScript in NPM_DEPENDENCIES.

brian428:
If that's the case, I wonder why it works fine when included in the dev build, but fails in the prod build.

sprell:
Likely due to the concatenation and minification for the prod build. For dev they are all separate script tags. Regardless, I manually commented out the "use strict" in relfect.js any my prob build works fine now. I know that's not a good log term solution, but it gets me past the issue for now. I could not simply remove the two shims altogether from the prod build, because I need them for IE to work properly.

atticae:
Hmm, maybe all "use strict" should be removed during bundling then?

mpetkov:
Looks like it is related to #818

[...]

PR #885, opened 2016-05-16
Add gulp-replace package and use it to strip the global 'use strict' …

brian428:
…added by reflect-metadata (#802).


brian428:
One minor change from what I had posted in the issue thread: this only strips the `'use strict'` preceding `var Reflect`. This way any other uses of `'use strict'` remain in place.

PR #889, opened 2016-05-17
Configurable client folder

troyanskiy:
With that it's possible to keep different clients in the project


troyanskiy:
Wow... It was fast :)
Thanks.

PR #886, opened 2016-05-17
feat(http): add sample http service calls

mgechev:
Fix #839

PR #892, opened 2016-05-17
Lazy Template Load

troyanskiy:
Added new flag for prod builder.
It's possible to choose include html templates into js or not.
Templates are included by default.


troyanskiy:
Tag name is LAZY_TEMPLATE
Or use -- --lazy-template to set

tools/config/seed.config.ts
```diff
@@ -79,6 +79,14 @@ export class SeedConfig {
   APP_BASE = argv['base'] || '/';
 
   /**
+   * The flag for the lazy template loader option of the application.
+   * Per default the option is not set, but can be set by the `--lazy-template`
+   * flag when running `npm start`.
+   * @type {boolean}
+   */
+  LAZY_TEMPLATE = argv['lazy-template'];
```
mgechev: Lets call this option: `INLINE_TEMPLATES`. `LAZY_TEMPLATE` could be a bit misleading because of the lazy-loading feature of the router.

mgechev:
LGTM, I added a single comment only. What do you think about my suggestion?

troyanskiy:
So if `INLINE_TEMPLATES` is `true`, then all templates will be included into JS file, otherwise not.
And default value is true.
Is that correct?
But than, if someone does not want to include templates is it possible to have something like `npm run build.prod -- --inline-templates=false` or `npm run build.prod -- --inline-templates=0`?

ludohenin:
@troyanskiy 
> Is that correct?

looks like it is.

mgechev:
`npm run build.prod -- --inline-templates false` seems fine.

troyanskiy:
Looks like AppVeyor test fails because of that merge https://github.com/mgechev/angular2-seed/pull/886

tools/tasks/seed/build.js.prod.ts
```diff
@@ -1,8 +1,9 @@
 import * as gulp from 'gulp';
 import * as gulpLoadPlugins from 'gulp-load-plugins';
 import { join } from 'path';
+import * as merge from 'merge-stream';
 
-import { TMP_DIR, TOOLS_DIR } from '../../config';
+import { TMP_DIR, TOOLS_DIR, INLINE_TEMPLATES, BOOTSTRAP_DIR, PROD_DEST } from '../../config';
```
TheDonDope: It'd be great if you could update the imports to conform to the styleguide :)

https://angular.io/styleguide#!#coding-conventions
> Import Line Spacing
> Style 03-06
> Do list destructured imported assets alphabetically.

ludohenin:
@troyanskiy Woud you rebase from master? there's conflicts.

troyanskiy:
@ludohenin Conflicts has been resolved

PR #916, opened 2016-05-23
chore(devDeps): Dependency Maintenance

### LINKED ISSUE #915: chore(devDeps): Update dev dependencies
**Major**

typings to 1.0.4
_Breaking Changes_ - https://github.com/typings/typings/releases/tag/v1.0.0

**Minor**

tslint to 3.10.2
Enhancements & new rules / rule options

**Patch**

ts-node to 0.7.2

**Associated Fixes**

PR is currently configured for the "new" typings install paths. The option exists to go back to main / browser if necessary though I can't see a good reason why we should go that route.

d3viant0ne:
Updates Typings Major Version

**Contains Breaking Changes** - You have to delete your typings folder and reinstall

Read "Updating from 0.x" section [here](https://github.com/typings/typings)

Resolves #915


Shyam-Chen:
@d3viant0ne 

Is it going to change?

https://github.com/mgechev/angular2-seed/blob/3bafd1851e097ac8d6148e956764864e567ad4aa/tools/tasks/seed/build.js.dev.ts#L18

d3viant0ne:
@Shyam-Chen - Missed that one, so yes that needs to be updated. That said, you would expect that to break and it doesn't which is something I need to look at.

PR #924, opened 2016-05-25
fix(tsconfig): remove typings

Shyam-Chen:


TheDonDope:
Hi @Shyam-Chen, why do we need to remove the typings exclusion? What is the current "faulty" behaviour with this exclusion?

d3viant0ne:
@Shyam-Chen - As far as I know, those are excluded purposefully in the tsconfig that covers the tools.

The app tsconfig on the other hand doesn't exclude as you would expect.

//cc @mgechev @ludohenin

ludohenin:
@d3viant0ne yep you're right. Not sure what would happen if not excluded, haven't tried.
It's not excluded in the client, but the compiler doesn't see it either (hence the `/src/client/typings.d.ts` file), as the project root is different. Typings are passed to the compiler in the tasks themselves.

mgechev:
Should be safe because the `tools` are using `typings/globals`. Not sure if it is necessary?

Shyam-Chen:
This is not necessary, but I think it is good to remove it.
`tools` using `main.d.ts` (Typings v0.x)

Issue #864, opened 2016-05-12
Advanced production build


mgechev:
Currently our production build is quite basic and produces only two bundles:

- shims
- libs + app

Which means that we're not taking advantage of:

- lazy loading
- offline template compilation
- tree-shaking

## Proposal

Incrementally improve the production build by:

### Step 1

[ ] - Include offline template precompilation
[ ] - Introduce tree-shaking with rollup.js

### Step 2

[ ] - Introduce bundling by bounded context (lazy-loading boundary) strategy

hermagrini:
Would Critical Path CSS be something interesting to generate as well?

phiphou:
As I said in #860 , application (only the code you/we wrote), vendors (angular2, rxjs and potential additional vendors like jquery, lodash...), and polyfills (reflect-metadata, zone, es6-shim,...) life-cycles aren't the same most of time. Keeping them in separate bundles seems to me a good practice. I'll be interested in knowing your thoughts about this point 

I also agree lazy loading would be a great feature. It would  obviously augment the amount of generated bundles, but I don't see any major reason not to working this way.

Bigous:
Lazy loading and hot loading are amazing. We could create a pattern ... I though, when it was changed, that components that started with + in it's directory would be lazy loaded only when needed. I like this pattern, what do you think?

jlberrocal:
@Bigous i'm using the + for indicate that is a lazy loaded component, for development seems to work fine, but for production there is still only 2 JS files, one with the shims and a big one with the vendors and the app itself

Bigous:
Yes thats why I think we could create a name pattern for that. Bundle the components in other files, like AngularClass did with the webpack starter kit would be a nice one and would resolve the step 2. Unfortunately I've got almost no time during the past 3 months to help with the project, but the idea is still very valid.

mgechev:
[Here's a blog post I wrote](http://blog.mgechev.com/2016/06/26/tree-shaking-angular2-production-build-rollup-javascript/) about an experiment I did around the production build efficiency.

A working "Hello world!" example with precompilation + tree-shaking, can be found [here](https://github.com/mgechev/angular2-ngc-rollup-build).

mgechev:
In addition - I'll be playing with the production build to introduce:

- Brotli
- Google Clojure Compiler

These tools should decrease the bundle size to around 20k.

Bigous:
[...]

PR #955, opened 2016-06-02
feature : configurable lazy loaded modules

taguan:
- Prod build now uses systemjs at runtime
- Prod build extracts libraries in a separate bundle
- New gulp task : build.modules, it builds the lazy modules that can be  configured in project config
- Lazy modules are loaded through system js and a virtual component configured in the routes

This pull-request is related to improvements proposed here : https://github.com/mgechev/angular2-seed/issues/864

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -50,6 +50,7 @@ gulp.task('build.prod', (done: any) =>
               'build.js.prod',
               'build.bundles',
               'build.bundles.app',
+              'build.modules',
               'build.index.prod',
               done));
 
--- a/src/client/app/+about/about.component.e2e-spec.ts
+++ b/src/client/app/+about/about.component.e2e-spec.ts
@@ -2,7 +2,7 @@ describe('About', () => {
 
   beforeEach( () => {
     browser.get('/');
-    element.all(by.css('nav > a')).get(1).click();
+    element.all(by.css('nav > a')).get(2).click();
   });
 
   it('should have correct feature heading', () => {
--- a/src/client/app/app.component.ts
+++ b/src/client/app/app.component.ts
@@ -2,13 +2,15 @@ import { Component } from '@angular/core';
 import { ROUTER_DIRECTIVES, Routes } from '@angular/router';
 import { HTTP_PROVIDERS} from '@angular/http';
 
-import { AboutComponent } from './+about/index';
 import { HomeComponent } from './+home/index';
+import { AboutComponent } from './+about/index';
 import { NameListService, NavbarComponent, ToolbarComponent } from './shared/index';
+import { componentProxyFactory } from './component_proxy_factory';
 
 /**
  * This class represents the main application component. Within the @Routes annotation is the configuration of the
- * applications routes, configuring the paths for the lazy loaded components (HomeComponent, AboutComponent).
+ * applications routes, configuring the paths for the components (HomeComponent, AboutComponent) and the lazy loaded
+ * component (LazyComponent).
  */
 @Component({
   moduleId: module.id,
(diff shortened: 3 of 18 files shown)


taguan:
I have the feeling that the system js configuration should be refreshed, for example by removing defaultJSExtensions which is deprecated, but I didn't want to mix multiple purposes to this PR.

TheDonDope:
Hi @taguan, first of all: 👏  👏  👏  nice work!
Could you provide a separate PR for the problem you described here:

> I have the feeling that the system js configuration should be refreshed, for example by removing defaultJSExtensions which is deprecated, but I didn't want to mix multiple purposes to this PR.

That'd be awesome!

Second, i have a few more questions:

How is this change related towards the angular2-rc.1 router? The rc1 router is currently not documented yet, so is this change of how to lazy load components the preferred way to implement this functionality?
Is this a SystemJS specific implementation? The Styleguide states here: [Styleguide Style 04-12 & Style 04-13](https://angular.io/docs/ts/latest/guide/style-guide.html#!#style-04-12):
image

image

Does your implementation suffice to that rules? Since the .rc1 Router is not yet documented ([Router Interface](https://angular.io/docs/ts/latest/api/router/index/Router-interface.html)) i have not seen a reference implementation for it.

Thank you again on that awesome work of yours! 🎉

taguan:
Hi @TheDonDope 
For the systemjs improvement, I will make a PR later if I have the time, it is not critical at all.

Regarding those rules, I have to admit I am a bit lost with the definition of lazy loading. Comments on the existing files explains that +home and +about are lazy loaded, but in my understanding, they just use the default router configuration. 
My lazy loading goes beyond that, it actually downloads the "js" only on demand.

I am not an angular expert, in my own project I am not even yet migrated to the new router, so I really don't know where my code falls regarding conventions.

TheDonDope:
Hi @taguan,
>  Regarding those rules, I have to admit I am a bit lost with the definition of lazy loading.

If you have a look at the screenshot where it states:

```typescript
// AVOID: app/app.component.ts
import { HeroesComponent } from './+heroes';
```

Your implementation satisfies this rule, since: https://github.com/taguan/angular2-seed/blob/77fcaa744c917fb1daebd35cf7e618fdc68adc50/src/client/app/app.component.ts#L33 here you really **do** lazy load on demand. You do not have an ```import``` on ```LazyModule```. So, if i understand correctly, this then is the right implementation of this rule. Or maybe better to say **one** correct implementation for it, which uses `SystemJS`. 

Regarding the router: If you take a look at the [Angular Weekly Meeting Notes](https://docs.google.com/document/d/150lerb1LmNLuau_a_EznPV1I1UHMTbEl61t4hZ7ZpS0/edit?pref=2&pli=1) you can see, that they are currently still at work with the router, so we may see another round of changes for rc2 (which is slated for Monday, 2016-06-06).

// cc @mgechev @ludohenin @NathanWalker @Shyam-Chen

taguan:
Ok, nice to read :-)

I am quite confident changes to the router won't break too much the code, because it uses very few features from it...

mgechev:
@taguan looks great!

I believe in the next release the core team will introduce asynchronous routing, which solve similar problem with less typing. On the other hand, you really did an awesome job. What do you think of adding a wiki page with this virtual proxy you did?

taguan:
I'll do that!
I don't think asynchronous routing will help with the way js are bundled in the prod build, I think this is the most interesting part of my contribution. The virtual proxy will probably be replaced by the feature you mention.

taguan:
@mgechev I finished implementing that lazy loading feature in my own project. Actually, the bundling works very well, but not the virtual component, because it does not work with nested routes (it expects some annotations on the virtual component it cannot find).
I am still using the deprecated router in my project, and I made it work perfectly with the AsyncRoute (actually, very few code change compared to this PR).

I propose I wait for Async implementation in the new router and some documentation and they I update this PR with implementation using the async feature of the router.

mgechev:
@taguan sounds good. I'm still using fork of the deprecated router myself, that I described here http://blog.mgechev.com/2016/05/21/angular2-router-implementing-missing-resolve-feature-deprecated-defer/.

PR #968, opened 2016-06-05
chore(config): introduce computed properties

### LINKED ISSUE #823: Cannot easily update computed config members in ProjectConfig
If you set APP_SRC inside ProjectConfig you don't get .css files injected into index.html.

I guess the reason is that properties like CSS_SRC and APP_ASSETS are calculated using old/default value of APP_SRC.

I can change APP_SRC inside SeedConfig as a workaround.

mgechev:
Fix #823

### DIFF
--- a/tools/config/project.config.ts
+++ b/tools/config/project.config.ts
@@ -3,6 +3,8 @@ import { join } from 'path';
 import { SeedConfig } from './seed.config';
 import { InjectableDependency } from './seed.config.interfaces';
 
+import { Overwrite } from '../utils/seed/overwrite_config';
+
 /**
  * This class extends the basic seed configuration, allowing for project specific overrides. A few examples can be found
  * below.
@@ -10,6 +12,7 @@ import { InjectableDependency } from './seed.config.interfaces';
 export class ProjectConfig extends SeedConfig {
 
   PROJECT_TASKS_DIR = join(process.cwd(), this.TOOLS_DIR, 'tasks', 'project');
+  @Overwrite() NPM_DEPENDENCIES: InjectableDependency[];
 
   constructor() {
     super();
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -12,6 +12,10 @@ export const ENVIRONMENTS: Environments = {
   PRODUCTION: 'prod'
 };
 
+const warnInvalidSet = (prop: string) => {
+  console.warn('You cannot set the property ' + prop);
+};
+
 /**
  * This class represents the basic configuration of the seed.
  * It provides the following:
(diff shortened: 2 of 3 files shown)


TheDonDope:
Hi!

looking good so far! How does this affect an override in `project.config.ts` ? For example here: https://github.com/mgechev/angular2-seed/blob/master/tools/config/project.config.ts#L24

I guess this code needs an update too, doesn't it?

mgechev:
Yeah, that is why the build fails. I'll fix it later today.

mgechev:
Lets discuss the latest proposed changes since they introduce different syntax for overwriting config properties.

Also, if you have ideas for clean and less verbose definition of the property setters which is going to keep the compiler quite let me know.

TheDonDope:
Hi! Thanks for the updates :)

I don't know if this is possible in TypeScript but wouldn't it be easer to **not** provide a `set` methode in `seed.config.ts` and instead override the `get` method in `project.config.ts` like this?

```TypeScript
// project.config.ts

get NPM_DEPENDENCIES: InjectableDependency[] {
   const seedDependencies = super.NPM_DEPENDENCIES;
   return seedDependencies.concat(additional_deps);
}
```

I do not know if it is possible to override a `get` method this way, but that would have been my initial naive approach to this.

mgechev:
Since we are using ts-node for gulpfile.ts and tsconfig which is independent from the project build config we can eventually have this. Unfortunately we need to drop all versions of node that don't support es6...
May be we can stick to this approach and improve it a bit.

tools/config/project.config.ts
```diff
@@ -3,13 +3,16 @@ import { join } from 'path';
 import { SeedConfig } from './seed.config';
 import { InjectableDependency } from './seed.config.interfaces';
 
+import {Overwrite} from '../utils/seed/overwrite_config';
```
Shyam-Chen: add whitespace?

```ts
import { Overwrite } from '../utils/seed/overwrite_config';
```

tools/config/seed.config.ts
```diff
@@ -285,136 +479,166 @@ export class SeedConfig {
...
-    'ie_mob >= 10',
-    'ff >= 30',
-    'chrome >= 34',
-    'safari >= 7',
-    'opera >= 23',
-    'ios >= 7',
-    'android >= 4.4',
-    'bb >= 10'
-  ];
+  get BROWSER_LIST() {
```
Shyam-Chen: add type?
```ts
get BROWSER_LIST(): string[]  {
```

mgechev: I think this is fine. We don't have implicit any and the compiler can easily perform type inference.

tools/config/project.config.ts
mgechev: Fixed.

TheDonDope:
Hey @mgechev, do you know which node versions would be affected (in terms of needing to be dropped because they do not support es6)? Anything < 5.x?
I am a bit on the fence here, though i can certainly understand the desire to deliver backwards compatibility. 
The PR definitely solves the overriding problem, but the code to fix this feels a little verbose (the need to provide the `set` methods to keep the compiler quiet). This (https://github.com/mgechev/angular2-seed/pull/968#issuecomment-223831937) feels a bit cleaner. But i guess we cannot have both (at least not right now).

To sum it up: For me the PR looks fine to merge, even though i wished for something a bit sleeker :)

Thank you @mgechev for tackling the issue! 🎉

mgechev:
Lets think of a better solution. I'm not huge fen of the one above tool.

TheDonDope:
One alternate solution i could think of would be to rewrite the properties to actual methods, like this:

before:
```TypeScript
  get PORT() {
    return argv['port'] || 5555;
  }
```

after:
```TypeScript
  getPort() {
    return argv['port'] || 5555;
  }
```

**Pros**:
- No need for for a `set` method
- No need for a `@Overwrite` annotation
- `project.config.ts` can easily override the methods, while still being able to access the `super` implementation

**Cons**:
- More work to rewrite the current code (Code that uses the current properties)

Thoughts on that?

ludohenin:
I was also thinking that we could wrap the computed config members into a `init` method and call that one at the end of `project.config` constructor.

```ts
class SeedConfig {
  //...
  init() {
    this.COMPUTED_MEMBER =  `${whatever}/blah`;
  }
}
```

__Pros__
* easy to refactor
* no breaking changes (I think)

__Cons__
* Overriding computed members will have to be done after the `init()` is invoked (convention)
* ...

mgechev:
The methods approach seems quite verbose too but we don't have to complicate the configuration by having additional conventions.

The `init` method approach seems simple. I don't like the convention but it doesn't seem like a big deal to ask people defile property values before `init()`.

The current approach has disadvantage that we have useless setters and we add extra `@Overwrite` decorator. The decorator doesn't seem like a big deal because we'd have already provided an example. The setters seem annoying. What I like in this approach is that we still have computed properties. We can change value of a base (non-computed) property runtime and everything is still going to work. But the entire approach here seems quite hacky.

I don't love any of these three options.

mgechev:
I'll think about a better solution over the weekend.

lukaszluczak:
@mgechev: Have you had a chance to think about better solution? (some weekends has passed ;) )

lifaon74:
Personnally (and propably for many windows user), we really need to find a solution !

Why :
- build.prod is not working with external libraries like ng2-translate, moment, etc... WE need to add them into SYSTEM_CONFIG_DEV.path, directly into seed.config.ts because of computation.
- some attributes like `APP_SRC` is used to compute other attributes into seed.config.ts, so we can't edit it properly into project.config !

The consequences: everytime I pull ng2-seedn (i I do it frequently because I love your work), I have to merge the seed.config file and take care of any changes :(

PR #999, opened 2016-06-16
Update to RC2

troyanskiy:
Angular library update to rc2


tools/config/seed.config.ts
```diff
@@ -299,13 +299,13 @@ export class SeedConfig {
...
     paths: {
       [this.BOOTSTRAP_MODULE]: `${this.APP_BASE}${this.BOOTSTRAP_MODULE}`,
-      '@angular/core': `${this.APP_BASE}node_modules/@angular/core/core.umd.js`,
-      '@angular/common': `${this.APP_BASE}node_modules/@angular/common/common.umd.js`,
-      '@angular/compiler': `${this.APP_BASE}node_modules/@angular/compiler/compiler.umd.js`,
-      '@angular/http': `${this.APP_BASE}node_modules/@angular/http/http.umd.js`,
-      '@angular/router': `${this.APP_BASE}node_modules/@angular/router/router.umd.js`,
-      '@angular/platform-browser': `${this.APP_BASE}node_modules/@angular/platform-browser/platform-browser.umd.js`,
-      '@angular/platform-browser-dynamic': `${this.APP_BASE}node_modules/@angular/platform-browser-dynamic/platform-browser-dynamic.umd.js`,
+      '@angular/core': `${this.APP_BASE}node_modules/@angular/core/index.js`,
```
mgechev: No bundles in rc.2?

mgechev: I'd say that the path should be `./bundles/core.umd.js`.

troyanskiy: Well...
I don'r really understand what does that mean. I just updated my project with rc2 and saw that project can't find ****.umd.js files. So I checked in the @angular/module folder and now there is everywhere index.js file
So I have updated seed.config.ts file and it works now for my project.

package.json
```diff
@@ -115,13 +115,13 @@
...
-    "@angular/platform-browser": "2.0.0-rc.1",
-    "@angular/platform-browser-dynamic": "2.0.0-rc.1",
-    "@angular/router": "2.0.0-rc.1",
+    "@angular/common": "2.0.0-rc.2",
+    "@angular/compiler": "2.0.0-rc.2",
+    "@angular/core": "2.0.0-rc.2",
+    "@angular/http": "2.0.0-rc.2",
+    "@angular/platform-browser": "2.0.0-rc.2",
+    "@angular/platform-browser-dynamic": "2.0.0-rc.2",
+    "@angular/router": "2.0.0-rc.2",
```
antonybudianto7: it should be 3.0.0-alpha.6, 2.0.0 is deprecated

troyanskiy: Wow... Thanks. Checking that.

tools/config/seed.config.ts
mgechev: The initial load time is faster. It loads the entire bundle instead of the angular modules one my one.

troyanskiy: Ahh... I see now...
Checking that.
Thank you!

package.json
```diff
@@ -115,13 +115,13 @@
...
-    "@angular/platform-browser": "2.0.0-rc.1",
-    "@angular/platform-browser-dynamic": "2.0.0-rc.1",
-    "@angular/router": "2.0.0-rc.1",
+    "@angular/common": "2.0.0-rc.2",
+    "@angular/compiler": "2.0.0-rc.2",
+    "@angular/core": "2.0.0-rc.2",
+    "@angular/http": "2.0.0-rc.2",
+    "@angular/platform-browser": "2.0.0-rc.2",
+    "@angular/platform-browser-dynamic": "2.0.0-rc.2",
+    "@angular/router": "3.0.0-alpha.3",
```
TheDonDope: Needs to be 3.0.0-alpha.**6**

troyanskiy: Oops

Issue #970, opened 2016-06-06
Protractor tests pass locally, but not in Travis CI


mraible:
**Steps to reproduce and a minimal demo of the problem**

I created a simple angular2 application using angular2-seed as a starting point. You can find it at https://github.com/mraible/angular2-tutorial. As part of creating this project, I wrote a tutorial on testing that you can find [here](https://github.com/mraible/angular2-tutorial/blob/master/TESTING.adoc). 

To reproduce the problem, fork my project and enable your repo in Travis CI. Then start a build. You should see an error like the following:

https://travis-ci.org/mraible/angular2-tutorial/jobs/135517549#L1190

**Current behavior**

Running `npm run webdriver-start`, `npm run serve.e2e` and `npm run e2e` passes locally, but [fails in Travis CI](https://travis-ci.org/mraible/angular2-tutorial/jobs/135517549#L1190). I believe this is because directly accessing URLs doesn't work in Travis.  For example, here's one of the tests:

```js
describe('Search', () => {

  beforeEach(() => {
    browser.get('/search');
  });

  it('should have an input and search button', () => {
    expect(element(by.css('sd-app sd-search form input')).isPresent()).toEqual(true);
    expect(element(by.css('sd-app sd-search form button')).isPresent()).toEqual(true);
  });

  it('should allow searching', () => {
    let searchButton = element(by.css('button'));
    let searchBox = element(by.css('input'));
    searchBox.sendKeys('M');
    searchButton.click().then(() => {
      // doesn't work as expected - results in 0
      //expect(element.all(by.repeater('person of searchResults')).count()).toEqual(3);
      var list = element.all(by.css('sd-search table tbody tr'));
      expect(list.count()).toBe(3);
    });
  });
});
```

Do my tests need to be re-written so they access the component by clicking on a link from the homepage, or is it possible to make this work in Travis CI like it does locally?

mgechev:
@Shyam-Chen would you take a look?

mraible:
I was able to workaround this issue by [navigating to each component](https://github.com/mraible/angular2-tutorial/commit/11fa5220d5264abfdc675323ca526e228e5204a5). However, I'd would love know if it's possible to just tweak some configuration to make this work.

Shyam-Chen:
Hi @mraible 

```ts
browser.get('/search');  // There can not be obtained directly URLs
```

harmandeol:
[...]

PR #1005, opened 2016-06-17
hotfix(e2e): fix travis broken tests

Shyam-Chen:
#970


harmandeol:
thanks Shyam-Chen. I just saw that.

Issue #985, opened 2016-06-12
Clean dev dependencies


mgechev:
From our initial goal to keep the seed minimalistic, it's set of dependencies grew over the last a couple of months. This made the entire installation process much longer, bringing a couple of extra plugins which may or may not be used by many.

I'm opening this issue with suggestion to drop:

- `audoprefixer`
- `colorguard`
- `doiuse`
- `gulp-clean-css`
- `gulp-postcss`
- `gulp-progeny`
- `gulp-sass`
- `gulp-sass-lint`
- `karma-phantomjs-launcher` (we already run tests in chrome)
- `postcss-reporter`
-` stylelint`
- `stylelint-config-standard`

I'd love to hear more thoughts. Do you really use all/any of these plugins or you prefer to keep the installation process lighter and get rid of them (or some of them)?

Note that there's a version of the seed which includes more functionality and respectively [more dependencies here](https://github.com/NathanWalker/angular2-seed-advanced).

// cc @ludohenin @d3viant0ne @TheDonDope @NathanWalker @Shyam-Chen

ckapilla:
The leaner the better! I personally would be happy to kiss all those good bye.

mgechev:
For reference:

Original seed:

```
$ time npm install
...
real    3m59.031s
user    1m33.492s
sys     0m22.507s
```

Minimalistic seed:

```
$ time npm install
...
real    2m46.400s
user    1m14.654s
sys     0m18.166s
```

corvinusy:
Availability of 2 variants would be fine.

mgechev:
Supporting three versions of the seed (minimalistic, the current one and the advanced seed of @NathanWalker) seems like a lot of effort, that I'm not sure is worth it. The minimalistic seed could be turned into the current version by adding Sass and css linting.

Shyam-Chen:
I would like to use `rucksack-css` instead of `autoprefixer`.

TheDonDope:
Hi!
I am a little on the fence about this topic. While i do support the strive for the least amount of dependencies, i think we shouldn't just try to throw out dependencies for lesser install time sake.

For example:

- `gulp-progeny`
- `gulp-sass`
- `gulp-sass-lint`

Those dependencies got included when @hank-ehly inlcuded the sass build option. People were happy to be able to use sass by a simply enabling a configuration flag. If we were to drop those, we would in consequence need to revert this build option too. 

[...]

PR #1016, opened 2016-06-19
Add support for stripping comments from .json files.

alantreadway:
Use-case for this is ng2-translate, where having comments on JSON
data files allows us to provide context for translations beyond the
data needed for the app (having comments at runtime in .json files
causes parsing failures).


tools/tasks/seed/build.assets.prod.ts
```diff
@@ -29,5 +31,6 @@ export = () => {
     '!' + join(ASSETS_SRC, '**', '*.js')
   ].concat(TEMP_FILES.map((p) => { return '!' + p; })))
     .pipe(onlyDirs(es))
+    .pipe(gulpIf((file) => ENABLE_JSON_COMMENT_STRIPPING && file.path.match(/.*.json$/), strip()))
```
alantreadway: This approach is quite specific to .json files, if you would prefer I could make the configuration a whitelist of file extensions - did not see any other examples like this in the existing config options though so kept it more specific.

mgechev:
Thanks for the PR! In this issue https://github.com/mgechev/angular2-seed/issues/985 we're discussing how to drop part of the development dependencies. Although the functionality you've implemented is definitely quite useful I prefer to not merge it with master yet in order to keep the set of dev dependencies manageable.

PR #1024, opened 2016-06-22
Upgrade to angular rc.3

juristr:
Updates the dependencies to the latest RC.3. I also adjusted the routes since with the `alpha7` of the new router, paths starting with a `/` are no more allowed.

Juri


mgechev:
Awesome! Thanks for the PR!

PR #1026, opened 2016-06-22
feat(config): add app environment specific config

mgechev:
Environment-specific configuration. Doesn't add extra dependencies and is quite handy.

Requested by @jbouzekri.


jbouzekri:
Nice ;)

Bigous:
Very clever.

PR #1032, opened 2016-06-23
chore: gulp-util instead of chalk

Shyam-Chen:


TheDonDope:
Looks good to me @Shyam-Chen ! This would further help to reduce the amount of dependencies :)
@mgechev Thoughts?

mgechev:
LGTM as well.

Issue #1034, opened 2016-06-23
The build.prod task should fail on TS error?


Sjiep:
Currently the `build.prod` will succeed, even when there are TS errors.

I can understand that this is desired behavior on `build.dev` since the watch would otherwise continuously break. However, to me it makes sense that when running the `build.prod` it should fail the build. (Since we have the `build.prod` running in a Jenkins CI)

@mgechev Is there an interest in including this functionality? I can make a PR for it.

mgechev:
Yes, please do.

PR #1037, opened 2016-06-23
build.prod task now fails on TS error

Sjiep:
Currently the build.prod will succeed, even when there are TS errors.

I can understand that this is desired behavior on build.dev since the watch would otherwise continuously break. However, to me it makes sense that when running the build.prod it should fail the build. (Since we have the build.prod running in a Jenkins CI)

### DIFF
--- a/tools/tasks/seed/build.js.prod.ts
+++ b/tools/tasks/seed/build.js.prod.ts
@@ -27,7 +27,11 @@ function buildTS() {
   let result = gulp.src(src)
     .pipe(plugins.plumber())
     .pipe(INLINE_TEMPLATES ? plugins.inlineNg2Template(INLINE_OPTIONS) : plugins.util.noop())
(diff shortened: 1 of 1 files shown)


TheDonDope:
According to #1034 looks good to me!

vincentpalita:
Hello, 

I don't agree with you on this because you could have typings error due to wrong typings from library you would include such as angular2-material.
For example they have the following typing error for now more than 1 month:
```/Users/vincent/Workspace/biznet.io/app/aphrodite/node_modules/@angular2-material/core/gestures/MdGestureConfig.d.ts(4,40): error TS2304: Cannot find name 'HammerManager'.
```
See: https://github.com/angular/material2/issues/797
But though, the library does run well.
With this now, the seed won't build at all.

Not sure this is something really "wanted".

PR #1044, opened 2016-06-25
chore: gulp instead of gulp-watch

Shyam-Chen:


TheDonDope:
Hi @Shyam-Chen, looking good to me :) And again one less dependency :)

mgechev:
I think there were some issues with gulp and that's why with @ludohenin we switched to gulp-watch.

ludohenin:
@mgechev I confirm that was the reason.

PR #1047, opened 2016-06-25
Clean index.html

TheDonDope:
Removes the workaround for the undefined module bundle in SystemJS
introduced by
https://github.com/mgechev/angular2-seed/commit/d7b41374b5a54fce90d9ca51e927e42c047f609d.

@ludohenin Do we still need this? I removed it and everything still worked well.

### DIFF
--- a/src/client/index.html
+++ b/src/client/index.html
@@ -14,11 +14,6 @@
 
   <sd-app>Loading...</sd-app>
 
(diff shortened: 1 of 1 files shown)


TheDonDope:
Ok, judging from the test it is indeed still needed :D

PR #1055, opened 2016-06-28
Make Stylus compiles down to CSS automatically.

13hun:
### New feature
Add gulp-stylus@2.5.0 into package.json.
Add gulpstylus.ts in tools/tasks/seed.
Make Stylus compiles down to CSS automatically.

### How to use
When you need edit the Stylus, you can edit it in the components files directly.
Use " *.component.styl " instead of " *.component.css "

### E.G.
Components often have four files (e.g. *.html, *.styl, *.ts, and *.spec.ts) 
>home.component.ts | html | styl | spec.ts


TheDonDope:
Hi @13hun, thank you for the PR! However, since the seed is aimed to provide a quite minimalistic setup i would rather see this as a new article in the wiki on how to integrate Stylus support.

@mgechev  Thoughts?

mgechev:
@13hun thanks for the effort! I agree with @TheDonDope that the purpose of the seed is to provide a minimalistic starter for development of Angular apps.

Lets close the PR for now.

PR #1061, opened 2016-06-30
Fixed a typo

daixtrose:
I believe an 'r' is missing


mgechev:
Thanks for the catch!

PR #1066, opened 2016-06-30
fix 'an/a' typo's  

ckapilla:
I hope you don't mind this extremely trivial PR, it is my first one, so I want to make sure I have the process down before trying anything more substantial.  Here is an [explanation of the change: ](http://www.everythingenglishblog.com/?p=541)

> Also consider words starting with “u.” Some such words make a “y” or consonant sound (e.g., unicorn, unique, and ukulele); in such instances, you should use the article “a” (e.g., a unicorn, a unique store, and a ukulele).

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -75,10 +75,10 @@ The seed provides the following utilities:
 
 | Filename               | Description |
 | :--------------------- | :---------- |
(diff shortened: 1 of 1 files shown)


mgechev:
Thanks for the catch!

Issue #973, opened 2016-06-07
importing rxjs operator timeInterval produce error "unexpected token <" in the browser


elpddev:
This is related ti #784 I think.

I tried:
```
import 'rxjs/add/operator/timeInterval';  
```

Which gives the following error in the browser:

```
(index):89 Error: SyntaxError: Unexpected token <
	    at Object.eval (http://localhost:5555/app/shared/scroll-tracker/scroll-tracker.service.js:13:1)
	    at eval (http://localhost:5555/app/shared/scroll-tracker/scroll-tracker.service.js:36:4)
	    at eval (http://localhost:5555/app/shared/scroll-tracker/scroll-tracker.service.js:37:3)
	Evaluating http://localhost:5555/node_modules/rxjs/add/operator/timeInterval
	Evaluating http://localhost:5555/app/shared/scroll-tracker/scroll-tracker.service.js
	Evaluating http://localhost:5555/app/app.component.js
	Evaluating http://localhost:5555/app/main.js
	Error loading http://localhost:5555/app/main.js "Report this error at https://github.com/mgechev/angular2-seed/issues"
```

1. When I try other operators, it works.
2. The timeInterval files exists in the node module rxjs add folder

I temporarily fixed the problem by commenting out the rxjs path and packages in the seed file as was mantioned in the related issue and it works. But I don't think this is a solution. 
```
paths: {
      [this.BOOTSTRAP_MODULE]: `${this.APP_BASE}${this.BOOTSTRAP_MODULE}`,
      //'rxjs/*': `${this.APP_BASE}node_modules/rxjs/*`,
      'app/*': `/app/*`,       
      '*': `${this.APP_BASE}node_modules/*`
    },
   packages: {                 
      //rxjs: { defaultExtension: false }
    }
```

mgechev:
You have entered a wrong path to the operator. The browser (by default) loads `index.html` which is executed as JavaScript by SystemJS and you get this error of unexpected token.

Make sure your import is correct.

nosachamos:
@mgechev path as indeed correct, and the file exists. I'm not sure why one must comment out those configs, but without that only certain operators can be actually added (such has map, filter, etc). You can try on any project - `timeInterval` cannot be added until these configs are commented out even though the file is there and the extension is .js just like the other operator patch files.

nosachamos:
It turns out the fix is to change the default extension for the rxjs package to `js`:

```
    packages: {
[...]

Issue #784, opened 2016-04-20
what is the right way to add rxjs operator? such as exhaustMap


goldenbearkin:
as subject. Anybody can help? i have been solving this for two days. Many thanks!

mgechev:
`import "rxjs/add/operator/exhaustMap";`

goldenbearkin:
i have already tried it, still doesnt work. Do a need to config the project.config.ts / see.config.ts?

mgechev:
No, you don't. Take a look at `rxjs/add/operator` and see if there's such file, ending with .js.

goldenbearkin:
yes. it has. but the broswer keeps yelling GET http://localhost:5555/rxjs/add/operator/exhaustMap.js 404 (Not Found)

mgechev:
I have a bunch of operators imported this way in my project:

```ts
import 'rxjs/add/operator/filter';
import 'rxjs/add/operator/map';
import 'rxjs/add/operator/combineAll';
import 'rxjs/add/operator/takeLast';

```

goldenbearkin:
```javascript
protected SYSTEM_CONFIG_DEV = {
        defaultJSExtensions: true,
        packageConfigPaths: [`${this.APP_BASE}node_modules/*/package.json`, `${this.APP_BASE}node_modules/@angular2-material/*/package.json`],
        paths: {
            [this.BOOTSTRAP_MODULE]: `${this.APP_BASE}${this.BOOTSTRAP_MODULE}`,
            'angular2/*': `${this.APP_BASE}angular2/*`,
            // 'rxjs/*': `${this.APP_BASE}rxjs/*`,
            'app/*': `/app/*`,
            '*': `${this.APP_BASE}node_modules/*`
        },
        packages: {
            angular2: {defaultExtension: false}
            // rxjs: {defaultExtension: false}
        },
    };
```
It works after comment out the rxjs related stuff. why is the rxjs without the node_modules prefix?

nosachamos:
@mgechev I'm not sure yet whats happening but @goldenbearkin has a point. The operators you import do work with the default seed configs, but some do not. I had lots of operators imported just like you, but then I tried to import `timeInterval` and it failed. I checked and it exists, path is correct, extension is .js just like the other operators, but it cannot be added until these configs are removed.

nosachamos:
It turns out the fix is to set the default extension for the rxjs package to `js`:

```
    packages: {
      rxjs: { defaultExtension: 'js' }
    }
```

PR #1087, opened 2016-07-05
fix(config): rxjs package's defaultExtension was set to true, but should be js instead

nosachamos:
Allows to import certain rxjs operators such as 'timeInterval' and 'exhaustMap'. Fixes https://github.com/mgechev/angular2-seed/issues/973 and https://github.com/mgechev/angular2-seed/issues/784


mgechev:
Thanks for the catch!

nosachamos:
Anytime! :+1:

Issue #1086, opened 2016-07-05
Sass support doesn't seem to work with unit tests


variuse:
When enabling SCSS support in seed or project config like `ENABLE_SCSS = true;` and using *.scss files instead of *.css files for component styling, unit tests can't run properly since *.css files are still referenced in components.

So in the end I tried both options: 

Referencing *.css in component definition like this
`styleUrls: ['about.component.css']`
Result: 
Karma crashes because it can't find *.css files since I'm using *.scss in src folder, but app is running properly from dist

Referencing *.scss in component definition like this
`styleUrls: ['about.component.scss']`
Result: 
Karma and unit tests run properly, but app crashes since it of course can't find *.scss files in dist 

I'm not an expert with Karma and apologize in advance if I'm doing something terribly wrong ;)

mgechev:
@variuse thanks for the issue! Most likely you're not doing anything wrong.

@hank-ehly would you have some time to take a look at this one?

hank-ehly:
@variuse 

Thank you for your comment. I'll take a look.

PR #1091, opened 2016-07-06
fix(1086) silence gulp-inline-ng2-template css warnings

hank-ehly:
Related to issue #1086
- Execute build.html_css before build.js.test to satisfy karma
- If using scss, gulp-inline-ng2-template will be unable to find the css files
  it needs. Setting the 'supportNonExistentFiles' boolean to true ignores this warning.

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -59,6 +59,7 @@ gulp.task('build.test', (done: any) =>
   runSequence('clean.dev',
               'tslint',
               'build.assets.dev',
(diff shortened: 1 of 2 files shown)


hank-ehly:
@mgechev 

I'll come up with some tests for the sass build.

mgechev:
@hank-ehly nice! Thanks for the fix.

variuse:
@hank-ehly Works great, thanks a lot!

PR #1092, opened 2016-07-06
Add express compression for Prod and Unit Test Servers

vyakymenko:
Add express compression for `serve.prod`.
I think this make sense, show developers real size of application with gzip compression and loading time. It's very important when you make a visual testing with dev tools and using _throttling_.

Same thing I have add for `unit test` server to speed up unit test run.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "browser-sync": "^2.11.2",
     "codelyzer": "0.0.20",
     "colorguard": "^1.1.1",
+    "compression": "^1.6.2",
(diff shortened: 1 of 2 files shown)


mgechev:
Makes sense. Does compression supports brotli as well?

vyakymenko:
@mgechev , as I know for now only `deflate` and `gzip` is supported.

mgechev:
Yeah, here are more details https://github.com/expressjs/compression/issues/71.

PR #1094, opened 2016-07-06
fine tuned browserSync reload to inject css

giovannicandido:
Hi,
Instead of a full reload, browserSync now inject the css files.
But when the css ends with *.component* a full reload is performed, issues with shadow dom I suppose.

Tested with both css and sass

Another way to look at it, is that global css changes in main.css or main.sass get inject, component css get reloaded.

That improve dev experience a little bit.

Another good thing to do, is modify the watch task to not recompile typescript  on each file change, that way the performance is improved.

### DIFF
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -406,7 +406,7 @@ export class SeedConfig {
       port: this.PORT,
       startPath: this.APP_BASE,
       open: argv['b'] ? false : true,
-      injectChanges: false,
+      injectChanges: true,
       server: {
         baseDir: `${this.DIST_DIR}/empty/`,
         routes: {
--- a/tools/utils/seed/code_change_tools.ts
+++ b/tools/utils/seed/code_change_tools.ts
@@ -2,7 +2,7 @@ import * as browserSync from 'browser-sync';
 // import * as path from 'path';
 
 import { getPluginConfig } from '../../config';
-
+import * as path from 'path';
 /**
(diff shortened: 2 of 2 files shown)


mgechev:
Thanks for the PR and for the comment! Yeah, it does inject CSS but it is bad practice to have CSS files with global scope (like main.css).

If you include CSS injection, the files associated to the components will not cause refresh of the page and the changed styles won't be applied (they need to be processed by Angular), which is confusing. We already dropped one such change.

I suggest to not merge such feature for now, until we are not able to process the changed component styles with Angular, server-side and be able to inject the emulated-shadow dom CSS

giovannicandido:
I agree that global css should be avoid, but its not a bad practice.

A global main.css is much less friction for themes. Shadow doom is great to isolate the css for each component, but for the same reason is a headache to use something like boostrap and many others.

The pattern I'm following is to use a main.scss that pick include components from libraries like uikit, or bootstrap and customize its theme. The component has css only for its basic layout, sometimes no css at all. The theme can dynamically change without coupling each component. That way I can customize the css framework, with the advantage to evolve the look and feel for all application, and across applications.

The goal is to let component only with behaviors and structure, and adapt the view to the needs of the application. For that, we need to combine the two technics. 

In the case of this pull request, the injection only happens if is not a **.component.css**, so the behavior is as expected. Maybe a better way to differentiate components css from application css, like filtering by a specific folder, but following the angular conventions, it works :-)

Anyway, I will put on the wiki, or provide a flag to opt for this feature.

mgechev:
Information in the wiki sounds good!

Issue #1098, opened 2016-07-07
Proposal: secondary, typeless compile option


brian428:
I've been using this in my own projects, so I thought I would see if you think this worth considering for a pull request.

Basically, `build.js.dev.ts` can be configured to do an initial fully-typed TS compile, but then, it will do a typeless and cached compile for the next N compiler runs. N is configured as a project config value (`FULL_COMPILE_INTERVAL`). I have this set to 6, but it can be set to anything, including -1 which will always do a full compile. Finally, if a compile error is encountered, it will keep triggering full compile runs to execute until the error is eliminated.

In short, you get a fully typed compile every N runs, with typeless compiles in between. Obviously, this can speed things up quite a bit, while still maintaining tolerable compile-time type checking.

I can paste in or attach the code (it's not that much code), or do a pull request, but first I just thought I'd ask if this is something you think might be useful from a conceptual standpoint.

mgechev:
I really like the idea, especially the part of it of performing type checking each next execution of `build.js.dev` until the error is fixed.

Still, I'd love to get a few more opinions cc @ludohenin @d3viant0ne @NathanWalker @TheDonDope.

In my personal project, I've disabled type-checking completely and I relay on vim and tsu plugin to perform it for me.

brian428:
Right. The main issue I've run into with relying on the IDE to do the type checking is that they generally don't constantly check the entire codebase, but only files that are open. So in my experience, it's not hard to make a change that could affect some other file, and the IDE doesn't flag it because that file isn't open. In that case, the compiler is the last line of defense. I didn't want to give that up, so I settled on the above approach as the best compromise I could come up with. :-)

mgechev:
Yes, this is an issue. We handle this using the production build/running tests. Your suggestion seems like an appropriate addition to me.

ludohenin:
Sounds good to me as well.

TheDonDope:
Since there will still be a 'do-always-compile' option (with `FULL_COMIPLE_INTERVAL = -1`), i do like the idea too. Would that also be the default option, so that we won't break with the current behaviour?

brian428:
Right, my assumption would be to have this off by default, and let people override the value in their project config if they choose to. (I also realized that I probably don't need to make -1 force full compiles, when 0 should work just as well.)

PR #1099, opened 2016-07-08
Feature/typeless compile

brian428:
Here's the typeless compile option mentioned in #1098.

A few quick notes:

- For some reason, when I use the typeless interval option, I'm occasionally seeing this gulp-typescript message:
```
Could not find input file C:\Developer\Projects\Personal\angular2-seed\src\client\________________empty.ts. This is probably an issue of gulp-typescript.
Please report it at https://github.com/ivogabe/gulp-typescript/issues
Debug information:
sourceRoot = "C:/Developer/Projects/Personal/angular2-seed/src/client/"
sources = ["________________empty.ts"]
```
I believe it's actually an issue with gulp-typescript, possibly caused by having two TS projects in use. If you test this out and also see the above message, you'll have to decide whether we can live with that message popping up.

- The use of `plugins.cached()` is only used if the typeless interval is turned off (e.g. `0`). I'm not positive, but I think that using the cached plugin while alternating between compile runs could wind up having the typed run miss files that were changed during the typeless runs, allowing errors to potentially sneak through uncaught. Hopefully this makes sense, let me know if it doesn't.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "gulp-sourcemaps": "2.0.0-alpha",
     "gulp-template": "^4.0.0",
     "gulp-tslint": "^4.3.3",
-    "gulp-typescript": "~2.13.4",
+    "gulp-typescript": "~2.13.6",
     "gulp-uglify": "^1.5.3",
     "gulp-util": "^3.0.7",
     "gulp-watch": "^4.3.5",
--- a/tools/config/project.config.ts
+++ b/tools/config/project.config.ts
@@ -13,6 +13,9 @@ export class ProjectConfig extends SeedConfig {
   constructor() {
     super();
     // this.APP_TITLE = 'Put name of your app here';
+    
+    /* Enable typeless compiler runs (faster) between typed compiler runs. */
+    this.TYPED_COMPILE_INTERVAL = 3;
 
     // Add `NPM` third-party libraries to be injected/bundled.
     this.NPM_DEPENDENCIES = [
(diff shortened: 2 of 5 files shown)


brian428:
By the way, the gulp-typescript `empty.ts` message seems to be a known issue: https://github.com/ivogabe/gulp-typescript/issues/342

brian428:
I also noticed that the gulp-typescript folks seem to think this is fixed in 2.13.5.

brian428:
I updated gulp-typescript to 2.13.6 and it seems to resolve the earlier "missing input file" message.

One thing I did discover is that the use of gulp-cached in the current master seems to be a problem. Gulp-cached seems to work by only passing changed files downstream. But this means that compile errors that affect other (currently unchanged) files will not be caught. 

Consider a simple example, like changing the name of a class. **That** class may still compile fine, but **other files** that reference the old class won't trigger a compile error because they have not changed (and thus are omitted by gulp-cached).

You may need to think about this and decide what tradeoff is acceptable here. Sure, using gulp-cached speeds the typed compile up. The question is whether the risk of having errors slip through is worth the speed increase. Perhaps this new TYPED_COMPILE_INTERVAL is a sufficient solution, meaning the typed compile will be slower but more thorough, while the more numerous typeless compiles will always be much faster?

For now, I've removed use of the cached plugin during the typed compiles. Let me know what you think when you have time to ponder this.

ckapilla:
>> The question is whether the risk of having errors slip through is worth the speed increase.

 

I can’t imagine anything would be worth the risk of having the build process miss errors; I think you were correct to turn it off. Thanks for working through all that!

brian428:
Thought I'd check on the status of this. If you've just been busy, that's fine. I just wanted to make sure you didn't see some problem with it that wasn't relayed. Thanks.

brian428:
Done!

PR #1122, opened 2016-07-15
Fix relative path on build.prod

### LINKED ISSUE #1040: Problem or Bug in build.prod  to sub-systems (APP_BASE)
Hi everyone,

I'm with @jvitor83 using "windows 10 and IIS" in the company where we work and we go through problems with seed due to sub-systems.

My suggestion so that other people using the same SO, can solve.

I have the following scenario:

Main application: www.site.com
Another application: www.site.com/subsite/

Problems found:

1 Any change in the "APP_BASE" in "SeedConfig class" modified the addresses of files *.js and *.css, getting always pointing to a subdomain more than I had

 1.1 - APP_BASE = argv['base'] || '/Subsite'; Generated the following url: www.site.com/subsite/#/subsite/ - She had no problems, but it is not correct.

2 Another option was to change the addresses of the script via files, as found here on an issue "npm run build.prod -- --env=prod --app-env=build.dev --dest=subsite --base=/subsite/ "and change the 
  2.1 This would be the best option because the project would be with APP_BASE pointing to my "subsite" and define "base url" as '/', making this option had the following problem.
2.1.1 Access Url: www.site.com/subsite/# was changed to www.site.com/#/subsite/ being impossible to access the functionality when it was accessed directly (Save to favorites), he tried to access a route Main application of "www.site.com/#" where not met and redirected to the page 404.

Solution:

The only solution we found after this test was because the idea of ​​@JMSantos94 in #755. Even so we would have to always manually change what was unfeasible

At that moment changed again the variable APP_BASE = argv ['base'] || '' <- No parameter.

It worked properly, without presenting any of the above reported problems.

I believe it can solve many of the problems with sub-systems that can appear in Windows.

Thanks!

### LINKED ISSUE #1107: prod build fails with extra fonts/images - concatCss issue
I am adding extra icons/images/fonts in `main.css`.
It all works fine in dev but fails in prod build.

Looking at the built `main.css` I found that the paths to my images/fonts has been transformed/rebased in the build process.
Paths like `../assets/icons/refreshIcon.png`, get converted to `../../src/client/assets/icons/refreshIcon.png`, and thus fail in production.

I looked at the `build.html_css.ts` task and found out that the path transform is done by a `concatCss` step, as it tries to rebase all relative urls.

So, I copied that task to `tools/tasks/project` and changed the call to `concatCss` to added an option to turn off relative url rebasing.

The call now looks like:
```
function processExternalCss() {
  return getExternalCssStream()
    .pipe(plugins.postcss(processors))
    .pipe(isProd ? plugins.concatCss(CSS_PROD_BUNDLE,{rebaseUrls: false}) : plugins.util.noop())
    .pipe(isProd ? cleanCss() : plugins.util.noop())
    .pipe(gulp.dest(CSS_DEST));
}
```

And everything works as expected, all relative urls are preserved in my .css files and resources are loaded as expected. (ps: I used advanced seed and the change also works in electron)

Is that the right solution to the problem? Should that be changed in the seed?
Or is there another way to tackle the problem?

jvitor83:
This closes #1040, and closes #1107

Apparently it is possible now to run `npm run build.prod` and publish the generated `dist/prod` to a subsite like `http://www.site.com/subsite/` and every related path should be ok.

### DIFF
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -67,7 +67,8 @@ export class SeedConfig {
 
   /**
    * The path for the base of the application at runtime.
-   * The default path is `/`, which can be overriden by the `--base` flag when running `npm start`.
+   * The default path is based on the environment ('/' for development and '' for production),
+   * which can be overriden by the `--base` flag when running `npm start`.
    * @type {string}
    */
   APP_BASE = argv['base'] || (this.ENV === ENVIRONMENTS.DEVELOPMENT ? '/' : '');
@@ -435,6 +436,18 @@ export class SeedConfig {
     'gulp-sass': {
       includePaths: ['./node_modules/']
     },
+
+    /**
+     * The options to pass to gulp-concat-css
+     * Reference: https://github.com/mariocasciaro/gulp-concat-css
+     * @type {object}
+     */
+    'gulp-concat-css': {
+      targetFile: this.CSS_PROD_BUNDLE,
+      options: {
+        rebaseUrls: false
+      }
+    }
   };
 
   /**
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -70,7 +70,7 @@ export class SeedConfig {
    * The default path is `/`, which can be overriden by the `--base` flag when running `npm start`.
    * @type {string}
    */
-  APP_BASE = argv['base'] || '/';
+  APP_BASE = argv['base'] || (this.ENV === ENVIRONMENTS.DEVELOPMENT ? '/' : '');
 
   /**
    * The base path of node modules.
--- a/tools/tasks/seed/build.html_css.ts
+++ b/tools/tasks/seed/build.html_css.ts
@@ -10,7 +10,6 @@ import {
   APP_SRC,
   BROWSER_LIST,
   CSS_DEST,
-  CSS_PROD_BUNDLE,
   CSS_SRC,
   DEPENDENCIES,
   ENABLE_SCSS,
@@ -21,6 +20,7 @@ import {
 
 const plugins = <any>gulpLoadPlugins();
 const cleanCss = require('gulp-clean-css');
+const gulpConcatCssConfig = getPluginConfig('gulp-concat-css');
 
 const processors = [
   autoprefixer({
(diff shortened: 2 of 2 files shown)


tools/tasks/seed/build.html_css.ts
```diff
@@ -97,7 +97,7 @@ function processExternalStylesheets() {
  */
 function processAllExternalStylesheets() {
   return merge(getExternalCssStream(), getExternalScssStream())
-    .pipe(isProd ? plugins.concatCss(CSS_PROD_BUNDLE) : plugins.util.noop())
+    .pipe(isProd ? plugins.concatCss(CSS_PROD_BUNDLE,{rebaseUrls: false}) : plugins.util.noop())
```
mgechev: Would you move rebaseUrl: false to pluginConfig in seed.config.ts?

tools/tasks/seed/build.html_css.ts
```diff
@@ -143,7 +143,7 @@ function getExternalScss() {
 function processExternalCss() {
   return getExternalCssStream()
     .pipe(plugins.postcss(processors))
-    .pipe(isProd ? plugins.concatCss(CSS_PROD_BUNDLE) : plugins.util.noop())
+    .pipe(isProd ? plugins.concatCss(CSS_PROD_BUNDLE,{rebaseUrls: false}) : plugins.util.noop())
```
mgechev: Same here.

mgechev:
LGTM. Would you address the comments?

tools/config/seed.config.ts
```diff
@@ -70,7 +70,7 @@ export class SeedConfig {
    * The default path is `/`, which can be overriden by the `--base` flag when running `npm start`.
    * @type {string}
    */
-  APP_BASE = argv['base'] || '/';
+  APP_BASE = argv['base'] || (this.ENV === ENVIRONMENTS.DEVELOPMENT ? '/' : '');
```
TheDonDope: Is it confirmable, that this change will not only "fix" the subsite issue for deploying on IIS, but also NOT breaking current deployments on non-IIS servers, or when no subsite structure is used?

rodolfocop: @TheDonDope 
Sim esta alteração feita soluciona os problemas de implantação no IIS, e  também não interfere nas implementações atuais em servidores não-IIS.

Issue #1048, opened 2016-06-27
Problem with watch and browser sync


martinsvb:
Hello,
may be I'm doing something wrong or have wrong expectation.
I run npm start and application is built in dev mode and localhost server started.
But after this actions changes aren't watched and propagate to the browser sync.
Please can you help me/describe problem?
Thanks
Martin

mgechev:
What is your OS, node.js, npm versions?

Did you make any changes in the seed or this occurs in fresh clone?

dbrezoev:
I think I have almost the same problem. Every time after npm start everything works just fine, but after I change something it throw me `Error: SyntaxError: Unexpected token < in JSON at position 0(…) "Report this error at https://github.com/mgechev/angular2-seed/issues"` I just added some classes (services) in the seed. I am on win 10, node 4.4.6, npm 2.15.5. I just tried with a fresh clone, and everything is working.

mgechev:
Would you share the content of your `tools/env/config.json`?

dbrezoev:
`{
  "dev": {
    "API": "https://demo.com"
  },
  "prod": {
    "API": "https://prod.com"
  },
  "staging": {
    "API": "https://staging.com"
  }
}
`

martinsvb:
I am on win 7, node 4.2.4, npm 3.10.2 (updated today, it was broken in v2 too)
It occurs in fresh clone.
I run browser at localhost:5555 (automatically opened), localhost:4000 is unavailable.
When I change for example home.component.html and save it, watch isn't processed.

martinsvb:
Hello again, it looks on problem with npm 3.10.2 v. On my older notebook with npm v2 version it works ok. Concretely I checkout repo, run "npm i" twice (maybe not needed) and "npm i typings" an it works ok.

martinsvb:
Anyway thank you for perfect seed! :-)

martinsvb:
Hi again, on other computer browser sync works with npm v3. So it isn't related to npm version. Probably some local problem in my notebook.

mgechev:
Ok, thanks. Lets close the issue then.

dbrezoev:
Hello again, I found out why my issue occurs. I use Visual Studio for both backend and frontend (typescript) and sometimes I misclick f5 for building the frontend instead of the backend. So it generates me all the `.js` files and browsersync stops working. After I discard these files everything is fine. I thought it would be nice if you know this.

jwbuitenhuis:
[...]

PR #1127, opened 2016-07-16
added templating to the build.js.test task

jwbuitenhuis:
When you run tests the config throws a parsing error.
The reason is that the templating wasn't activated for this flow yet.

Related to: https://github.com/mgechev/angular2-seed/issues/1048


mgechev:
Would you fix the build in order to merge?

njs50:
I was looking at this with regards to the coverage being broken. I think perhaps the build.js.test task should only be building the tests and the build.tests task should also run build.js.dev 

seems bad to be having two seperate paths for building the same dev js.

this would also allow the watchers to only rebuild js or tests depending on what had actually changed.

PR #1130, opened 2016-07-16
update some dependencies

sfabriece:
update some dependencies to reflect their latest versions


mgechev:
Awesome job thanks! Would you only revert the version of the `systemjs-builder` back? We used the previous one intentionally because with the newer one there were problems resolving dependencies.

sfabriece:
done, hope everything is ok now. :smiley:

package.json
```diff
@@ -35,81 +35,81 @@
...
-    "stylelint-config-standard": "^6.0.0",
+    "stylelint": "^7.0.2",
+    "stylelint-config-standard": "^11.0.0",
     "systemjs-builder": "0.15.14",
     "tildify": "^1.2.0",
     "tiny-lr": "^0.2.1",
-    "traceur": "^0.0.91",
-    "ts-node": "^0.7.3",
-    "tslint": "^3.10.2",
+    "traceur": "^0.0.111",
```
Shyam-Chen: I think this use of less than.

PR #1142, opened 2016-07-20
fix(coverage): resolve issues preventing coverage/tests running correctly

njs50:
here's a PR with what I did to get coverage working and the tests updated for RC4. 

after this update you should be able to run `npm test` and `npm run serve.coverage` and have it work as expected. karma should also be able to detect changes properly when running the watch task.

build updates:
update build.js.test task to only build the tests (different build paths invalidate the sourcemaps)
update gulp build.test task to only perform a clean on the initial build and to also run build.js.dev now
add a clean.coverage task and add it to the build.test task

test updates:
remove deprecated jasmine imports
add mock router/location strategy (until @angular provides one)
add default values to home.component so home component tests will pass.

not done:
figure out some way to run istanbul-remap whenever a karma run completes.

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -1,4 +1,5 @@
 import * as gulp from 'gulp';
+import * as util from 'gulp-util';
 import * as runSequence from 'run-sequence';
 
 import { PROJECT_TASKS_DIR, SEED_TASKS_DIR } from './tools/config';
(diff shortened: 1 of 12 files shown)


TheDonDope:
Hey Nick!

Thank you so much for this!! Great job!

njs50:
happy to help. some of it might need a little rearranging, but it seems like an okay starting point. 

I'm trying to get a thing hooked up to karma-coverage atm to trigger the remap, would be ideal if it could just sort itself out without us needing to run anything.

PR #1146, opened 2016-07-21
package.json dependencies update

vyakymenko:
* dependencies update.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "stream-series": "^0.1.1",
     "stylelint": "^7.0.2",
     "stylelint-config-standard": "^11.0.0",
-    "systemjs-builder": "0.15.23",
+    "systemjs-builder": "0.15.14",
     "tildify": "^1.2.0",
     "tiny-lr": "^0.2.1",
     "traceur": "^0.0.111",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -97,11 +97,11 @@
     "stream-series": "^0.1.1",
     "stylelint": "^7.0.2",
     "stylelint-config-standard": "^11.0.0",
-    "systemjs-builder": "0.15.14",
+    "systemjs-builder": "0.15.23",
```
mgechev: This version is locked intentionally.

package.json
```diff
@@ -122,8 +122,8 @@
     "@angular/router": "3.0.0-beta.2",
     "es6-module-loader": "^0.17.8",
     "core-js": "^2.4.0",
-    "rxjs": "5.0.0-beta.6",
-    "systemjs": "0.19.27",
+    "rxjs": "5.0.0-beta.10",
```
mgechev: This update causes the build failure.

vyakymenko: @mgechev , you right, i have reverted this change.

package.json
vyakymenko: @mgechev , my false, reverted.

vyakymenko:
@mgechev , re-test it, now everything ok.

AloeDream:
One of new packages causes this to fail. Had exact same problem with prod build when manually updated everything except rxjs to latest versions. Spent hours then rolled package.json back and prod build started performing well.

PR #1177, opened 2016-08-02
fix(makeTsProject): use proper tsconfig.json file

### LINKED ISSUE #1174: src/client/tsconfig.json is totally ignored by "gulp tasks" that compiles typescript
1) Clone repo
2) Change src/client/tsconcfig.json (for example change module: "system" or even module: "any_string")
3) Perform gulp build.dev
4) Inspect dist/dev/app/main.js and see that it produced the same as with module: "commonjs" option.

As i understand we needs two configs tsconfig.json - for tools and src/client/tsconfig.json for application.

mgechev:
Fix #1174

### DIFF
--- a/tools/utils/seed/tsproject.ts
+++ b/tools/utils/seed/tsproject.ts
@@ -1,4 +1,7 @@
 import * as gulpLoadPlugins from 'gulp-load-plugins';
+import { join } from 'path';
+
+import { APP_SRC } from '../../config';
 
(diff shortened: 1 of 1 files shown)


TheDonDope:
Looks good to me, will give it a test ride later today!

TheDonDope:
I just tested it locally on my machine, everything seems to still be fine :)
:shipit:

PR #1206, opened 2016-08-11
Fix for build.tools task.

dmitriyse:
Fix for Issue https://github.com/mgechev/angular2-seed/pull/1177 brokes build.tools task.

This is the fix.

### DIFF
--- a/tools/tasks/seed/build.js.tools.ts
+++ b/tools/tasks/seed/build.js.tools.ts
@@ -11,7 +11,7 @@ const plugins = <any>gulpLoadPlugins();
  * Executes the build process, transpiling the TypeScript files within the `tools` directory.
  */
 export = () => {
-  let tsProject = makeTsProject();
+  let tsProject = makeTsProject({}, true);
   let src = [
     'typings/index.d.ts',
     TOOLS_DIR + '/manual_typings/**/*.d.ts',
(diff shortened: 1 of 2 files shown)


dmitriyse:
Also I suggest to add some checks for all gulp task into CI builds.

dmitriyse:
Looks like machine was busy and timeout reached. Please restart build on travis-ci

mgechev:
@dmitriyse what error does it throw? I don't see the build failing.

dmitriyse:
Hi! The problem is next: I am using systemjs for src/client part. build.tools wrongly uses src/client/tsconfig.json and fails due to incompatibility with systemjs. So your will not reproduce the problem in a current master state. Unfotunately my project is commertial and I cannot share full compilable sources to reproduce.

mgechev:
It seems the problem is specific for your use case. I'd recommend you to overwrite the `build.js.tools` task in `tools/tasks/project`, which basically means that you need to copy/paste your version of the file to `tools/tasks/project`. Pushing such a change upstream doesn't seem required.

dmitriyse:
Hi, I agree this issue does not affects most of the angular2-seed users. But for those who Uses Visual Studio 2015 as FrontEnd development tool the only way to debug src/client/*.ts files is to switch from commonjs to systemjs. So the number of developers affected is probably more than hungdred. In a few words when /tsconfig.json and /src/client/tsconfig.json are different we have a problems. Probably not only VS 2015 developers are affected.
For me it does not matter any more, I implemented fix and go futher. But I spent for this problems ~ 6 hours. Other developers will also fill more ng2 pain.

Issue #1274, opened 2016-09-01
Update to RC.6


angelstam:
Angular 2 RC.6 is now available.

gvsdan:
i have a problem, if some of my external dependencies use the previous RC wont work ?

PR #1275, opened 2016-09-01
Angular 2.0.0 rc.6

Nightapes:
* update to angular 2.0.0 rc.6 (#1274)
* update to typescript 2.0.0
* update tests


src/client/app/about/about.component.spec.ts
```diff
@@ -1,12 +1,12 @@
-import { disableDeprecatedForms, provideForms } from '@angular/forms';
+import { FormsModule } from '@angular/forms';
 import { Component } from '@angular/core';
```
ludohenin: I don't think we need the forms in about module.

ludohenin:
Awesome!

src/client/app/about/about.component.spec.ts
Nightapes: Thanks for the hint, i removed it.

Issue #1305, opened 2016-09-06
npm run serve.prod -- --app appfoo


netstart:
Clone app to appfoo, 
execute 
npm run serve.prod  cause below problem:
![image](https://cloud.githubusercontent.com/assets/200232/18291240/371f6bf2-745d-11e6-859c-a0ac138f59da.png)

execute 
npm run serve.prod -- --app appfoo
or
npm start -- --app appfoo
cause the same problem

mgechev:
Are you using the latest version of the seed? There were related issues which I fixed yesterday and todai.

netstart:
Yes,  I cloned the repository a few minutes

![image](https://cloud.githubusercontent.com/assets/200232/18291480/3bd925ec-745e-11e6-93a3-5be12a325d14.png)

tinuF:
Might be related to https://github.com/mgechev/angular2-seed/issues/1300. If there's one ts error, serve.prod will fail even if tslint is commented out in gulpfile.

tinuF:
Might be related to https://github.com/mgechev/angular2-seed/issues/1300. If there's one ts error, serve.prod will fail even if tslint is commented out in gulpfile.

sasikumardr:
Line number 26 in build.js.prod make sure that ${BOOTSTRAP_FACTORY_PROD_MODULE}.ts not included in files to compile. Interesting to see it got picked up. 
Can you please check again what src files are considered in your local copy and master branch?

mgechev:
@netstart `build.js.prod` does not include `${BOOTSTRAP_FACTORY_PROD_MODULE}.ts` which is the only file which requires `app.module.ngfactory.ts` so most likely you're not using the latest version. Make sure you've recompiled your build files if you're on windows.

netstart:
Sorry, but no.

I Cloned the repository again, and we have problem yet

![image](https://cloud.githubusercontent.com/assets/200232/18348577/83b119de-75a2-11e6-98ab-0f8adf22a8d8.png)

![image](https://cloud.githubusercontent.com/assets/200232/18348607/b69b4bee-75a2-11e6-933a-a94e4e72e14c.png)

sasikumardr:
Can you plz paste your build.js.prod content here? Also seed.config.ts showing BOOTSTRAP_MODULE, BOOTSTRAP_PROD_MODULE values

netstart:
In the linux we have problem to.

[...]

PR #1312, opened 2016-09-08
Skip main-prod.ts in non-AoT build for multi-app scenario. Fix for #1305

sasikumardr:
In case of multi-app scenario, non AoT builds will skip only main app's main-prod.ts file. This commit will ensure that all main-prod.ts in APP_SRC will be skipped in transpiling process (build.js.dev, build.js.prod and build.js.e2e)


sasikumardr:
@mgechev this commit fixes non AoT builds. But we need to check about AoT builds. Does the current setup takes care of multi-app AoT builds?

mgechev:
It should as soon as some conventions are followed.

Issue #1269, opened 2016-08-31
endpoints with "." not working


Perezmarc:
**Steps to reproduce and a minimal demo of the problem**
http://localhost:5555/users/a.b

Want to access there. "a.b" is a username, but all I get is:

```
<html><head></head>
<body>Cannot GET /users/a.b
</body></html>
```

Any endpoint having a user without a "." char is working perfectly.

My Routes: 

```
export const routes: Routes = [
  { path: '', redirectTo: 'home', pathMatch: 'full', canActivate: [AuthGuard] },
  { path: 'login', component: LoginComponent },
  { path: 'home', component: HomeComponent, canActivate: [AuthGuard] },
  { path: 'discover', component: DiscoverComponent, canActivate: [AuthGuard] },
  { path: 'profile', component: ProfileComponent, canActivate: [AuthGuard] },
  { path: 'settings', component: SettingsComponent, canActivate: [AuthGuard] },
  { path: 'users', children: [
    {path: ':username', component: ProfileComponent, canActivate: [AuthChecker]}
  ]},
];
```

Thanks!

mgechev:
This is not an issue in angular2-seed. We're using the standard Angular 2 router. For further information take a look at the official [Angular 2 repo](https://github.com/angular/angular).

Perezmarc:
I'm using angular 2 router...

```
import { RouterModule } from '@angular/router';

@NgModule ( {
  imports: [CommonModule, BrowserModule, HttpModule, RouterModule.forRoot ( routes ), HomeModule, LoginModule,
    DiscoverModule, ProfileModule, SettingsModule, SharedModule.forRoot (),
```

This is the way I import it, 

version: "@angular/router": "3.0.0-beta.2",

Perezmarc:
For the record: I discovered is a grunt configuration issue, with the compiled version works as desired

mgechev:
In such case it's the seed :-). If you find the root cause, it'll be great if you can open a PR!

Perezmarc:
I would like to show some light to the issue, I have too much work to do yet before we release 😞 , need to focus... I don't need it in development, as long as it works good in production don't have the need to solve it 😃 .

sasikumardr:
[...]

PR #1330, opened 2016-09-13
handle endponts with '.' in dev environemnt (browser-sync middleware)…

sasikumardr:
Fix for #1269. Adding rewrite rules for connect-history-api-fallback middleware to serve static file as it is and disable dot checking


mgechev:
@sasikumardr awesome job! Let's wait for the build to pass and merge it.

PR #1349, opened 2016-09-17
replacing typings for @types

daniru:
Replacing old method to get type definitions by the new @types in Typescript 2.0.
[The Future of Declaration Files](https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/)


package.json
```diff
@@ -39,6 +39,38 @@
...
+    "@types/rimraf": "^0.0.27",
+    "@types/run-sequence": "^0.0.27",
+    "@types/selenium-webdriver": "^2.44.28",
+    "@types/systemjs": "^0.19.30",
+    "@types/through2": "^0.0.28",
+    "@types/vinyl": "^1.1.28",
+    "@types/vinyl-source-stream": "^0.0.27",
+    "@types/yargs": "^0.0.30",
+    "@types/zone.js": "^0.0.26",
     "async": "^2.0.0",
```
mgechev: Would you update all the dependencies for which you have installed `@types` to the same version as the typings themselves? For instance: `async` to be `^2.0.31` instead of `^2.0.0`.

daniru: As far as I know there is not such relation between the library and the @type.
Last version of async is [2.0.1](https://github.com/caolan/async) and the last version of the @types/async is 2.0.31.

I added all the last versions of the @types, abd they should be correspondent with the DefinitelyTyped repository that typings were importing.

PR #1365, opened 2016-09-21
chore: clean typings

Shyam-Chen:


.gitignore
```diff
@@ -22,8 +22,7 @@ build/Release
 # Dependency directory
 # Commenting this out is preferred by some people, see
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git-
-/node_modules/
-/typings/
+node_modules
```
mgechev: Would you keep typings? Some users may still have both in their workspace.

ZuSe: Question is if you want support both. Actually everyone should go for TypeScript 2 and the new system. I would get rid of the old stuff and force updates.

mgechev: By adding both typings and types in gitignore we don't support both, we just ignore both.

Issue #1366, opened 2016-09-21
Protractor typings


tgoyer:
Getting the following error for each of the defs in the Protractor index.d.ts running the "start" script:
`node_modules/@types/protractor/index.d.ts(22,27): error TS2305: Module 'webdriver' has no exported member 'IButton'.`

Brand new clone.  Just ran `npm install` and then `npm start`.  

Do I need to update anything on my box to get this to work or is there some breaking change to the type definitions?

Full error output:

[23:53:48] Starting 'build.js.dev'...
node_modules/@types/protractor/index.d.ts(22,27): error TS2305: Module 'webdriver' has no exported member 'IButton'.
node_modules/@types/protractor/index.d.ts(25,24): error TS2305: Module 'webdriver' has no exported member 'IKey'.
node_modules/@types/protractor/index.d.ts(28,45): error TS2339: Property 'Error' does not exist on type 'typeof error'.
node_modules/@types/protractor/index.d.ts(29,40): error TS2305: Module 'webdriver.error' has no exported member 'IErrorCode'.
node_modules/@types/protractor/index.d.ts(36,37): error TS2305: Module 'webdriver.logging' has no exported member 'IType'.
node_modules/@types/protractor/index.d.ts(37,38): error TS2305: Module 'webdriver.logging' has no exported member 'ILevelValues'.
node_modules/@types/protractor/index.d.ts(39,67): error TS2305: Module 'webdriver.logging' has no exported member 'ILevel'.
node_modules/@types/protractor/index.d.ts(40,67): error TS2305: Module 'webdriver.logging' has no exported member 'ILevel'.
node_modules/@types/protractor/index.d.ts(287,39): error TS2339: Property 'stacktrace' does not exist on type 'typeof webdriver'.
node_modules/@types/protractor/index.d.ts(288,42): error TS2339: Property 'stacktrace' does not exist on type 'typeof webdriver'.
node_modules/@types/protractor/index.d.ts(303,35): error TS2305: Module 'webdriver' has no exported member 'stacktrace'.
node_modules/@types/protractor/index.d.ts(340,55): error TS2305: Module 'webdriver' has no exported member 'Locator'.
node_modules/@types/protractor/index.d.ts(414,52): error TS2305: Module 'webdriver' has no exported member 'Locator'.
node_modules/@types/protractor/index.d.ts(462,53): error TS2305: Module 'webdriver' has no exported member 'Locator'.
[...]

PR #1369, opened 2016-09-21
Fix for Issue "#1366: Protractor typings"

markwhitfeld:
This pull request fixes issues mentioned in issue #1366 Protractor Typings.
Reversed the recent commit to the src/client/tsconfig.json which broke the build.
This was presenting itself as many errors of the form: "error TS2300: Duplicate identifier ..."
The 2.53 version of the @types/selenium-webdriver breaks compatibility with version 2.44.
This breaking change wrecked the build for protractor, so I pinned the version to 2.44.*.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "@types/protractor": "^1.5.18",
     "@types/rimraf": "^0.0.27",
     "@types/run-sequence": "^0.0.27",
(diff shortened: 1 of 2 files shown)


fedeguzman:
This fixed my project too! Thanks 👍

PR #1390, opened 2016-09-26
Build systemjs compatible rxjs bundle in order to reduce loading time in DEV mode

karlhaas:
see https://github.com/angular/angular/issues/9359

Reduces the number of initial HTTP requests in dev mode from 115 to 63. In a larger app this significantly reduces the initial loading time of the app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "gulp": "gulp",
     "karma": "karma",
     "karma.start": "karma start",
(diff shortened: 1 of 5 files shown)


mgechev:
LGTM. It'll be good if you add a comment that it's a temporal fix, related with https://github.com/angular/angular/issues/9359. This way we'll know to remove it once https://github.com/angular/angular/issues/9359 is closed.

karlhaas:
comment added

PR #1391, opened 2016-09-26
Removed package configuration and fixed loading of package.json files

karlhaas:
In SYSTEM_CONFIG_DEV the configuration of each angular bundle should not be necessary since the bundles are the main files of the node packages.

In SYSTEM_BUILDER_CONFIG loading of the package.json files didn't work because of the missing path `'node_modules/*': 'node_modules/*',`. Bundling then works fine without configuring each package.

### DIFF
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -320,6 +320,25 @@ export class SeedConfig {
     ],
     paths: {
       [this.BOOTSTRAP_MODULE]: `${this.APP_BASE}${this.BOOTSTRAP_MODULE}`,
+      '@angular/common': 'node_modules/@angular/common/bundles/common.umd.js',
+      '@angular/compiler': 'node_modules/@angular/compiler/bundles/compiler.umd.js',
+      '@angular/core': 'node_modules/@angular/core/bundles/core.umd.js',
+      '@angular/forms': 'node_modules/@angular/forms/bundles/forms.umd.js',
+      '@angular/http': 'node_modules/@angular/http/bundles/http.umd.js',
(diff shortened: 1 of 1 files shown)


mgechev:
I really like the simplification that you propose, however, the tests are failing. Most likely changing `SYSTEM_CONFIG_DEV` broke the testing configuration which prevents packages for being loaded. I can merge once the issue is fixed.

mgechev:
The Windows CI still fails. One of my future goals is to simplify the build configuration because I believe this is one of the places where we can achieve much better results, so I'm also interested in this. It'll be great if we can have only a single System config for both - the builder and for dev. At this point I haven't done any work in this direction because it seems error-prone and less flexible but it's definitely in the roadmap.

Let me know if you have any ideas about this, otherwise we can close the PR for now and come back to this improvement when we feel more confident in the solution.

karlhaas:
Yes, sorry for the inconvenience ... I reverted all changes except the change in SYSTEM_BUILDER_CONFIG.paths. I added `'node_modules/*': 'node_modules/*'`.

Without this change the builder fails to load the package.json files of node packages. I think that, because of `'*': 'node_modules/*'` it tries to load from `node_modules/node_modules`.

After this fix, the builder found additional packages without adding entries to packages. Because of that, I thought that we don't need the packages anymore and I tried to reduce the configuration even more. Sadly, I forgot the tests ...

Anyhow, currently it takes a lot of time to add additional packages (in our case lodash, moment, ng2-bootstrap, ng2-translate, ngrx, ...) and to keep the build system up-to-date.

mgechev:
No problems, thank you for trying to improve the usability of the starter! I agree that it's a bit verbose, I'm looking forward for the moment when we'll be able to reduce the configuration.

PR #1394, opened 2016-09-27
feat(codelyzer): add extra linting rules

mgechev:


TheDonDope:
Great stuff! Are `app.component` and `toolbar.component` the only two files which need an update to satisfy the new rules?

mgechev:
Uups, yes, it seems I have missed something locally.

PR #1397, opened 2016-09-27
Add all needed rxjs/ngrx operators to one central file

karlhaas:
RXJS/NGRX operators must be added to the application by importing them like ` import 'rxjs/add/operator/map';`. If, like in our application, operators are used a lot, they have to be imported often.

We would like to have one central place where all needed operators are added. Because those operators might be used in tests as well, their types definitions must be available in `build.js.test.ts`. For this purpose I import the operator definitions in a manual typings file.

Are there better solutions for this?

### DIFF
--- a/src/client/app/app.component.ts
+++ b/src/client/app/app.component.ts
@@ -1,5 +1,6 @@
 import { Component } from '@angular/core';
 import { Config} from './shared/index';
+import './operators';
 
 /**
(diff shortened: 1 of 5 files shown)


mgechev:
I think that with the current setup your PR makes change. If we introduce lazy loading in near future we'll have to refactor it.

mgechev:
Waiting for the CI to pass its checks to merge.

PR #1399, opened 2016-09-28
Moved karma reporters config from task to config for easier customization

karlhaas:
In order to keep the build tasks up-to-date with the seed project, it is easier to have all customization options in the config.


karlhaas:
Don't know why the test failed. Can you please re-run it?

PR #1406, opened 2016-09-29
Add "start.deving" task for continuous testing and live-reload of coverage and site

markwhitfeld:
We have found this task to be extremely useful when developing so we decided to contribute it back to the project.
What do you think?

I tried not to mess with any of the existing files (these tasks were just from our tools/tasks/project folder). You may want to rework some of the files if it does not match your overall design of the seed tasks.


TheDonDope:
Hey @markwhitfeld ,

Thank you very much! I think this is a great idea!
@mgechev what are your thoughts?

Looks good to me!

mgechev:
It was even more inconvenient before the ["test.watch"](https://github.com/mgechev/angular2-seed/blob/master/package.json#L15) task was added.

Yes, LGTM too!

markwhitfeld:
Great, thanks for the merge!
One thing I would like to point out though is the fact that I decided to exclude the index.ts and *.module.ts files from the coverage (this is different to the other coverage tasks in the seed) - see the commit [6b44626](https://github.com/mgechev/angular2-seed/pull/1406/commits/6b44626b784ab4446945216f6781c80642ed17fa) above.

I was getting frustrated that my project test coverage was looking a lot better than it actually was because of these files that end up getting tested as part of the scaffolding of the project/tests.

Maybe this is not something that everyone would want... maybe it is? What do you think?
Anyway, it is currently inconsistent with the other coverage tasks.
In retrospect I should have excluded this commit from the pull request and added it as a separate pull request so that a decision could be made separately. Maybe you would like to exclude that change? If you do so then I will open a separate pull request for it.

JunaidZA:
Excellent idea @markwhitfeld , excluding the scaffolding files gives a much more accurate code coverage figure. 
I've created a pull request (#1423) to also exclude route files, as they have the same skewing effect of the index and module files on test coverage percentage.

Another recommendation would be to rename the task to 'start.developing', 'deving' doesn't sound like a word, what are your thoughts?

PR #1413, opened 2016-10-02
fixed typo in contributing.md

robstoll:
this way it refers to the project the seed is used in and not to the seed project which makes more sense IMO


mgechev:
Thanks for the PR!

I'd prefer to keep it the way it is in order to not break links in case we restructure the repository and move `contributing.md` to `docs`, for instance. Would you update the PR with the first change only?

PR #1414, opened 2016-10-02
feat(banner): add 256 and plain banners

mgechev:
A while ago @ludohenin suggested to add a banner. I don't remember why we didn't do that at this time. I think having this will give a better identity of the project.

I added two banners:

1. Plain ascii one for terminals without 256 color support.
2. An advanced one, which includes the Angular logo.

`supports-color` is not an overhead because it is already required by some of the modules we use (we may install it twice in case different versions are required in the dependency graph, in which case we may need to download additional ~1k).

In this PR I'm also adding a warning in `seed.config.ts` which suggests to not edit this file.

Btw, I'm open for suggestions for better/more consistent banners + logo :-).


TheDonDope:
OH MY GOD I LOVE IT ❇️ 🎇 ✨

ludohenin:
awesome!

PR #1418, opened 2016-10-03
overridable tslint.json and enabled comments

robstoll:
- moved tslint.json to tools/config/seed.tslint.json
- tslint extends seed.tslint.json instead
- comments in tslint.json were not possible so far due to the require in seed.config.ts
- removed customRules() and CODELYZER_RULES from seed.config.ts they are no longer needed, gulp-tslint can detect the rules directory via tslint.json (adapted tslint.ts accordingly)


mgechev:
LGTM, I'm surprised tslint doesn't complain about the JavaScript comment in `tslint.json`.

robstoll:
Quite a nice feature in my opinion. This is always something I miss in json based configuration files. However, some IDEs might consider it as error. Thus, I am also fine if we remove the comment

mgechev:
I approved it, would you only rebase against master so I can merge.

PR #1428, opened 2016-10-05
fix(sass): not moving .sass files to dist/* folder

ruffiem:
The seed is already dealing with scss files but not with sass files. Since the functionality is already there to use sass files with the seed, it would be nice to handle them as well.


mgechev:
Thanks for the PR!

PR #1437, opened 2016-10-06
fix(build): allow start with differents environments

isidroamv:
Actually is not possible run the command `npm start -- --config-env prod`  because `npm script` executes  `"start": "gulp serve.dev --color"`

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -5,10 +5,18 @@ import * as runSequence from 'run-sequence';
 import Config from './tools/config';
 import { loadTasks } from './tools/utils';
 
+let ENV = Config.ENV || 'dev';
 
 loadTasks(Config.SEED_TASKS_DIR);
 loadTasks(Config.PROJECT_TASKS_DIR);
 
+
+// --------------
+// Serve env
+gulp.task('serve.env', (done: any) =>
+  runSequence('build.'+ENV,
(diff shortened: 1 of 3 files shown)


mgechev:
With `--config-env` you're choosing from the configuration variables that you want to be used (i.e. `--config-env prod` you'll be using the configuration for production). With `--env` you're setting the build - if you want it to be a development build or a production build (with AoT, bundling, minification, etc.)

With `npm start -- --config-env prod` you should be able to run the development build and dev server with production environment variables.

isidroamv:
I thought that with `npm start -- --config-env prod` I could serve a production environment. However is not possible execute that command with production variables because "gulp" is getting variables from `seed.config.ts ` and it takes `--env` argument instead of `--config-env`.

mgechev:
It is getting the environment (not build environment but environment configuration) from `--config-env`, here [it is](https://github.com/mgechev/angular2-seed/blob/master/tools/utils/seed/template_locals.ts#L25).

Does `npm start -- --config-env prod` outputs the production env variables for you?

isidroamv:
Yes, but I was using [this](https://github.com/mgechev/angular2-seed/blob/master/gulpfile.ts#L5) variable, I think `--config-env` is only available for build procedure.

PR #1442, opened 2016-10-08
feat(i18n): angular i18n native implementation

ruffiem:
- angular i18n native implementation for JiT processing
- ~~added systemjs-text-plugin to allow systemjs deal with XLIFF files~~
- application entry point at `app.through.ts` in `seed/tasks` folder
- `npm run i18n` task to extract content and create former messages.xlf
- translation files to be stored in `src/client/assets/locale`


mgechev:
@ruffiem I took only a brief look at your changes but it looks great so far! Thanks for the PR, I'll merge the it tonight or tomorrow morning.

ruffiem:
Thanks! Just saw there's something wrong with the production build, it breaks on "System.import is not a function" in `i18n.providers.ts`. 😠

ruffiem:
Working on the issue, hope to fix it by the end of the day. Does systemjs text plugin may cause such errors ?

mgechev:
Yes, in production we are excluding SystemJS from the bundle. I'd recommend you to use the Angular's http service instead. Otherwise it looks good to me!

ruffiem:
Great, thanks. I didn't know System.import was just an xhr :) I will refactor a little, run a few tests and push when everything seems correct.

mgechev:
@ruffiem another small thing - we also support AoT compilation so you'd have to update `main-prod.ts` because it has different bootstrap logic.

ruffiem:
Ok now it does work on both dev and prod but we have a bigger challenge : when scanning the code for i18n attributes, angular give a specific id based on the childnode. Thing is, with uglification the id is changing and the translation is lost. Can't think of anything "easy" to fix it right now.

mgechev:
The id is changed by uglify performing mangling? If so, we can disable mangling for build.prod and leave it only for build.prod.exp where we may not have such problem because of the AoT.

ruffiem:
Exactly.

Also, mangling the XLF is not possible since ids are hardcoded in the file.

I created an issue on the angular repo as they should rely on content instead of formatting IMO.

ruffiem:
I have updated the text content to avoid translation to break from the seed on prod, ids won't change but the text has to be written inline.

```html
<h1 i18n>Howdy!</h1>
```

not

```html
<h1 i18n>
     Howdy
!
</h1>
```

Also, TP from `main.ts` were causing `ng-xi18n` fn to break. It's fixed now.

mgechev:
@ruffiem you can also disable mangling, which should be fine.

ruffiem:
I have tried to disable mangling but the result was the same (except that the ids were changing but still not matching the ones from the XLF). Did you try too ?

mgechev:
@ruffiem I'll give it a try later today. I'm releasing beta.1 of codelyzer today with bug fixes.

mgechev:
I took a final look. It looks good, thanks for the effort one more time. My suggestion will be to keep it disabled by default because a lot of people will be confused by the bootstrap process.

Lets keep `main.ts` like this:

```ts
/**
 * Bootstraps the application and makes the ROUTER_PROVIDERS and the APP_BASE_HREF available to it.
 * @see https://angular.io/docs/ts/latest/api/platform-browser-dynamic/index/bootstrap-function.html
 */
import { enableProdMode } from '@angular/core';
// The browser platform with a compiler
import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
// Load i18n providers
// import { TranslationProviders } from './i18n.providers';

// The app module
import { AppModule } from './app.module';

if (String('<%= ENV %>') === 'prod') { enableProdMode(); }

// Compile and launch the module with i18n providers
// let TP = new TranslationProviders();
// TP.getTranslationFile().then((providers: any) => {
//   const options: any = { providers };
platformBrowserDynamic().bootstrapModule(AppModule/*, options*/);
// });
```

I've also removed the comments for the Service Workers. We can add a wiki article of how the i18n can be enabled (basically uncomment the `TranslationProviders` and the `xlf` file generation).

What do you think about this?

ruffiem:
Sounds great, I will make a wiki page about how to use i18n if you allow me to. And thanks to you for the seed, do not hesitate to tell me if you need help or something 👍

mgechev:
Sounds good! Looking for the updates in the PR and the wiki page, then I'll be ready to merge.

ruffiem:
Ok, should be fine now, I've also updated the package.json to match current master.

ruffiem:
Since i18n is an option, maybe we should create a few options in the `project.config.ts`. The language directory is hardcoded in the AoT task which is not a great option 👎  what do you think ?

src/client/app/i18n.providers.ts
```diff
@@ -0,0 +1,30 @@
...
+      xhr.onload = (data: any) => resolve(
+        [
+          { provide: TRANSLATIONS, useValue: data.target.response },
+          { provide: TRANSLATIONS_FORMAT, useValue: 'xlf' },
+          { provide: LOCALE_ID, useValue: locale }
+        ]
+      );
+      xhr.onerror = () => reject(noProviders);
+      xhr.send();
+    });
```
Bigous: I think we could use the Angular2 HTTP provider here instead of XMLHttpRequest.

src/client/app/home/home.component.html
```diff
@@ -1,6 +1,4 @@
-<p>
-  Howdy! Here's a list of awesome computer scientists. Do you know any others? Add to the list yourself.
-</p>
+<p>Howdy! Here's a list of awesome computer scientists. Do you know any others? Add to the list yourself.</p>
```
Bigous: Could you add some more examples on use (like how do we can translate numbers and variables and/or plurals)

src/client/app/i18n.providers.ts
ruffiem: I tried but it was resolving to `undefined` maybe I was missing something...

mgechev: @Bigous this service is invoked before bootstrap so I'm not sure if it is possible since the i18n functionality is part of the compilation process (in this case JiT).

mgechev:
@ruffiem good [wiki](https://github.com/mgechev/angular-seed/wiki/Enabling-Angular-i18n-native) page!

tools/tasks/seed/app.through.ts
```diff
@@ -0,0 +1 @@
+import '../../../src/client/app/main';
```
mgechev: @ruffiem this file doesn't seem used anywhere, right?

src/client/app/i18n.providers.ts
Bigous: @mgechev True. It should be a service where the Application module depends so... or keep it as it is :)

tools/tasks/seed/app.through.ts
ruffiem: @mgechev it's used by `ng-xi18n` as an entry point to the application. Since `ng-xi18n` relies on the `tsconfig.json` and `src` folder it excluded, I needed to find a way in 🙄

mgechev: I got it. Can you pass `src/client/tsconfig.json` as argument of `ng-xi18n`? For instance in ngc it'll be: `ngc -p src/client/tsconfig.json`.

ruffiem: There's no -p option with ng-xi18n yet but it's being discussed on the angular repo ([2201](https://github.com/angular/angular-cli/issues/2201)).

src/client/app/home/home.component.html
ruffiem: Yes, totally, but it seems like pluralization is not working well (especially 'other'). [plnkr](http://plnkr.co/edit/SxWt8hnnCaKNy5AbXjxI?p=preview)

tools/tasks/seed/app.through.ts
mgechev: Thanks! I added this as [last step](https://github.com/mgechev/angular-seed/wiki/Enabling-Angular-i18n-native) in order to not pollute the seed's tasks directory.

PR #1443, opened 2016-10-09
feat(tools): do not invoke tasks if not necessary

### LINKED ISSUE #1432: Speed-up the development build
## Problem

Recently, while using the seed, I had to create a task which execution takes 5s. It is for svg sprite creation and compression but it was triggered on any file change. This is a lot for dev build since it blocks everything else, so the livereload cannot refresh the browser until the execution of the slowest task completes. I thought what we can do and with the current stateless task management (which is great but a bit limiting) there was no right way to approach.

## Proposal

I was thinking that we can have a global state, which is mutated by `utils/seed/watch.ts`. So for instance:

```ts
export function watch(taskname: string) {
  return function () {
    let paths:string[]=[
      join(Config.APP_SRC,'**')
    ].concat(Config.TEMP_FILES.map((p) => { return '!'+p; }));

    plugins.watch(paths, (e: any) => {
      changedFiles.add(e.path);
      runSequence(taskname, () => {
        changedFiles.clean();
        notifyLiveReload(e);
      });
    });
  };
}
```

Later in the individual tasks, based on the file type we can check if any of the changed files requires their activation and can be skipped otherwise. For instance:

```ts
// build.js.dev
export = (done: any, filesChanged: string[]) => {
  if (hasTypeScriptFile(filesChanged)) {
    return gulp.src(...)
           .pipe(ts())
           .pipe(gulp.dest(...));
  } else {
    done();
  }
};
```

We can extend this even further by exporting not only a single function but a task object which processes only tasks of given type and/or has any other activation preconditions:

```ts
type FileType = RegExp;

interface ITask {
  precondition(files: string[]);
  processFiles?: FileType[];
  run(done: any): GulpStream | any;
}
```

```ts
const task: ITask = {
  precondition(files: string[]) {
    // decide if it should be activated
  },
  processFiles: [/\.ts$/],
  run(done) {
    // ...
  }
};
```

By default given task will be executed in all cases.

In order to keep the code DRYier, it'll be best to move the filtering logic in the task runner in `tasks_tools.ts`.

Basically it'll be a type of a chain of responsibility with inversion of control, because the tasks are not the ones which decide if the rest of the tasks in the chain should be executed.

mgechev:
In dev executes tasks only when a specific file type has changed.
Fix #1432.

### DIFF
--- a/tools/tasks/assets_task.ts
+++ b/tools/tasks/assets_task.ts
@@ -1,0 +1,10 @@
+import { Task } from './task';
+
+export abstract class AssetsTask extends Task {
+  shallRun(files: String[]) {
+    return files.reduce((a, f) => {
+      return a || (!f.endsWith('.css') && !f.endsWith('.sass') &&
+       !f.endsWith('.scss') && !f.endsWith('.ts'));
+    }, false);
+  }
+}
--- a/tools/tasks/css_task.ts
+++ b/tools/tasks/css_task.ts
@@ -1,0 +1,10 @@
+import { Task } from './task';
+
+export abstract class CssTask extends Task {
+
+  shallRun(files: String[]) {
+    return files.some(f =>
+      f.endsWith('.css') || f.endsWith('.sass') || f.endsWith('.scss'));
+  }
+
+}
--- a/tools/tasks/seed/build.assets.dev.ts
+++ b/tools/tasks/seed/build.assets.dev.ts
@@ -1,20 +1,25 @@
 import * as gulp from 'gulp';
 import { join } from 'path';
 
+import { AssetsTask } from '../assets_task';
 import Config from '../../config';
 
 /**
  * Executes the build process, copying the assets located in `src/client` over to the appropriate
  * `dist/dev` directory.
  */
-export = () => {
-  let paths: string[] = [
-    join(Config.APP_SRC, '**'),
-    '!' + join(Config.APP_SRC, '**', '*.ts'),
-    '!' + join(Config.APP_SRC, '**', '*.scss'),
-    '!' + join(Config.APP_SRC, '**', '*.sass')
-  ].concat(Config.TEMP_FILES.map((p) => { return '!' + p; }));
+export =
+  class BuildAssetsTask extends AssetsTask {
+    run() {
+      let paths: string[] = [
+	join(Config.APP_SRC, '**'),
+	'!' + join(Config.APP_SRC, '**', '*.ts'),
+	'!' + join(Config.APP_SRC, '**', '*.scss'),
+	'!' + join(Config.APP_SRC, '**', '*.sass')
+	    ].concat(Config.TEMP_FILES.map((p) => { return '!' + p; }));
 
-  return gulp.src(paths)
-    .pipe(gulp.dest(Config.APP_DEST));
-};
+      return gulp.src(paths)
+	.pipe(gulp.dest(Config.APP_DEST));
+    }
+  };
+
--- a/tools/tasks/seed/build.html_css.ts
+++ b/tools/tasks/seed/build.html_css.ts
@@ -7,6 +7,7 @@ import * as util from 'gulp-util';
 import { join } from 'path';
 
 import Config from '../../config';
+import { CssTask } from '../css_task';
 
 const plugins = <any>gulpLoadPlugins();
 const cleanCss = require('gulp-clean-css');
@@ -147,4 +148,14 @@ function processExternalCss() {
 /**
  * Executes the build process, processing the HTML and CSS files.
  */
-export = () => merge(processComponentStylesheets(), prepareTemplates(), processExternalStylesheets());
+export =
+  class BuildHtmlCss extends CssTask {
+
+    shallRun(files: String[]) {
+      return super.shallRun(files) || files.some(f => f.endsWith('.html'));
+    }
+
+    run() {
+      return merge(processComponentStylesheets(), prepareTemplates(), processExternalStylesheets());
+    }
+  };
--- a/tools/tasks/task.ts
+++ b/tools/tasks/task.ts
@@ -1,0 +1,25 @@
+/**
+ * Base class for all tasks.
+ */
+export abstract class Task {
+  /**
+   * Override this task if you want to implement some custom
+   * task activation mechanism. By default each task will be always executed.
+   *
+   * @param {string[]} files A list of files changed since the previous watch.
+   */
+  shallRun(files: string[]): boolean {
+    return true;
+  }
+
+  /**
+   * Implements your task behavior.
+   *
+   * @param {any} done A function which should be activated once your task completes.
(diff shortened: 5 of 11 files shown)


tools/utils/seed/tasks_tools.ts
```diff
@@ -14,6 +17,33 @@ export function loadTasks(path: string): void {
...
+	if (task.length > 0) {
+	  return task(done);
+	}
+
+	const taskReturnedValue = task();
+	if (isstream(taskReturnedValue)) {
+	  return taskReturnedValue;
+	}
+
+	done();
```
robstoll: ending here (indentation problem)

tools/utils/seed/tasks_tools.ts
```diff
@@ -14,6 +17,33 @@ export function loadTasks(path: string): void {
...
+      }
+    };
+  }
+  throw new Error(`${taskName} should be instance of the class
+Task, a function or a class which extends Task`);
```
robstoll: indentation problem as well

tools/utils/seed/watch.ts
```diff
@@ -17,8 +18,13 @@ export function watch(taskname: string) {
...
 
-    plugins.watch(paths, (e:any) =>
-      runSequence(taskname, () => notifyLiveReload(e))
-    );
+    plugins.watch(paths, (e: any) => {
+      changeFileManager.addFile(e.path);
+
+      runSequence(taskname, () => {
+	changeFileManager.clear();
+	notifyLiveReload(e);
```
robstoll: Indentation problem

tools/tasks/css_task.ts
```diff
@@ -0,0 +1,11 @@
+import { Task } from './task';
+
+export abstract class CssTask extends Task {
+
+  precondition(files: String[]) {
+    return files.reduce((a, f) => {
```
robstoll: couldn't all those reduce calls be simplified with `files.some(f => f.endsWith('.css') || ...)`?

tools/tasks/seed/build.assets.dev.ts
```diff
@@ -1,20 +1,24 @@
...
-    join(Config.APP_SRC, '**'),
-    '!' + join(Config.APP_SRC, '**', '*.ts'),
-    '!' + join(Config.APP_SRC, '**', '*.scss'),
-    '!' + join(Config.APP_SRC, '**', '*.sass')
-  ].concat(Config.TEMP_FILES.map((p) => { return '!' + p; }));
+export =
+  class BuildAssetsTask extends AssetsTask {
+    run() {
+      let paths: string[] = [
+	join(Config.APP_SRC, '**'),
```
robstoll: indentation problem

tools/tasks/seed/build.assets.dev.ts
```diff
@@ -1,20 +1,24 @@
...
+	'!' + join(Config.APP_SRC, '**', '*.ts'),
+	'!' + join(Config.APP_SRC, '**', '*.scss'),
+	'!' + join(Config.APP_SRC, '**', '*.sass')
+      ].concat(Config.TEMP_FILES.map((p) => { return '!' + p; }));
 
-  return gulp.src(paths)
-    .pipe(gulp.dest(Config.APP_DEST));
-};
+      return gulp.src(paths)
+	.pipe(gulp.dest(Config.APP_DEST));
```
robstoll: indentation problem

tools/tasks/seed/build.html_css.ts
```diff
@@ -147,4 +148,16 @@ function processExternalCss() {
 /**
  * Executes the build process, processing the HTML and CSS files.
  */
-export = () => merge(processComponentStylesheets(), prepareTemplates(), processExternalStylesheets());
+export =
+  class BuildHtmlCss extends CssTask {
+
+    precondition(files: String[]) {
+      return super.precondition(files) || files.reduce((a, f) => {
+	return a || f.endsWith('.html');
```
robstoll: indentation

tools/tasks/assets_task.ts
```diff
@@ -0,0 +1,10 @@
+import { Task } from './task';
+
+export abstract class AssetsTask extends Task {
+  precondition(files: String[]) {
+    return files.reduce((a, f) => {
+      return a || (!f.endsWith('.css') && !f.endsWith('.sass') &&
+		   !f.endsWith('.scss') && !f.endsWith('.ts'));
```
robstoll: I would indent this line once more

tools/tasks/task.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Base class for all tasks.
+ */
+export abstract class Task {
+  /**
+   * Override this task if you want to implement some custom
+   * task activation mechanism. By default each task will be always executed.
+   *
+   * @param {string[]} files A list of files changed since the previous watch.
+   */
+  precondition(files: string[]): boolean {
```
robstoll: I would name the method differently, something like `shallRun`, would make its usage clearer IMO 
```
if (task.shallRun(changeFileManager.lastChangedFiles)) {
  return task.run(done);
}
```

tools/tasks/task.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Base class for all tasks.
+ */
+export abstract class Task {
```
robstoll: You could provide another abstract class which uses a regular expression to test file endings. This way a subclass would only need to pass an array of file endings to this abstract class instead of overriding` precondition` (`shallRun` respectively, see below)

tools/utils/seed/tasks_tools.ts
```diff
@@ -14,6 +17,33 @@ export function loadTasks(path: string): void {
...
+  if (task instanceof Task) {
+    return task;
+  }
+  if (task.prototype && task.prototype instanceof Task) {
+    return new task();
+  }
+  if (typeof task === 'function') {
+    return new class AnonTask extends Task {
+      run(done: any) {
+	if (task.length > 0) {
```
robstoll: hm... here it looks good, not sure if it is a bug in the review view of github or if it is shown nicely here (the same for the other hints about indentation)

tools/tasks/task.ts
mgechev: Yes, I considered this idea [here](https://github.com/mgechev/angular-seed/issues/851#issuecomment-251251654). It might be a good idea but I'd prefer to keep the API minimalistic for now. Maybe in future will turn out that the most important precondition is not related to file type.

mgechev:
@robstoll thanks for the review! It seems vim messed the indentation up during the rebase I did. `shallRun` is more clear so I renamed it too.

tools/tasks/task.ts
Bigous: Maybe it's too late for this one, but I'll follow NG2 pattern and use a `canRun` or something like that. Sorry, I was out today.

PR #1450, opened 2016-10-10
WIP: feat(build): add tree-shaking with rollup

mgechev:
**Do not merge**.

Here's sample rollup.js build. Unfortunately, the bundling is not quite successful. Rollup includes two instances of rxjs and the operators imports patch the wrong `Observable`.


Bigous:
Hi @mgechev, You are working with both. Do you feel that rollup is better than Clojure or it's there because clojure is still implementing other things?

mgechev:
I thinkt that tsickle + closure can provide better results because of the option to run closure in advanced mode with preserved type annotations.

Periodically I am doing experiments with rollup and this time we are really close! However, even bundling the simple seed app involves some workarounds so I don't think we are there yet.

Shyam-Chen:
use [gulp-rollup](https://github.com/mcasimir/gulp-rollup)?

Shyam-Chen:
https://github.com/rangle/webpack-blog-code

Issue #1453, opened 2016-10-10
precompiled ts artifacts + gitignore?


njs50:
I ran into a weird issue with one of our dev's where they had precompiled ts in their /tools dir, where there is a gitignore that prevents them from showing up in `git status`. i.e `tools/config/project.ts` -> `tools/config/project.ts/project.js+project.js.map`

during the build it was using those old artifacts without recompiling them and causing issues (is this a tsconfig issue?).

i don't see any precompiling happening in the current version, so maybe that .gitignore should be removed so at least the old files show up?

lifaon74:
It's normal => https://github.com/mgechev/angular-seed/wiki/Speed-up-the-build-on-Windows
Files are pre-compiled. You'll need to execute `tsc --project tsconfig.json` everytime you change something into tools/. Personnaly I recommand not to precompile, avoiding to do it manually.

njs50:
this happened to a developer's build who doesn't ever touch the angular code on osx. not sure how those files would have been created.

seems like the sort of thing the clean should be removing before a prod.build if we are going to be ignoring them

mgechev:
If you run `tsc` in the root of the seed, all the TypeScript files in `tools` will be compiled to JavaScript. As @lifaon74 mentioned, in some cases this could be intentional because of the slow TypeScript compilation in Windows which slows down the entire build process.

It is also possible any editor or other tool that your coworker is using to have invoked the TypeScript compiler. If you're **sure** that you don't have any important JavaScript or sourcemaps in `tools`, you can run:

```shell
find . -type f -name '*.js.map' -delete && find . -type f -name '*.js' -delete
```

njs50:
it seems like a bad that a prod build is using old artifacts, and that none of the clean tasks will clear those out. I would expect a prod build to not use any existing artifacts, and be 100% repeatable.

Given the tendency of ide's to trick users into compiling their ts this seems like something that could potentially be quite confusing.

mgechev:
Yes, it is not recommended but brings a serious performance boost so it is the "last step" that you can take in order to speed-up the Windows build process.

njs50:
i don't think the speed matters for a prod build. maybe we need a task in the npm `build.prod` that checks to see if the compiled ts exists and then recompiles it if it does.

mgechev:
It brings improvement for both - development and production. On Windows it used to get really slow until your build gets started. 

[...]

PR #1456, opened 2016-10-11
improve precompiled tool handling

njs50:
following discussion in #1453

replaced existing build.tools with exec that replicate the command suggested in "speed up the build on windows" wiki page so results will be the same 

fixes issue where running the "windows command" then clean.tools breaks the system by updating the clean.tools task (previously wasn't clearing compiled gulpfile.ts).

added a check.tools command which checks if the tools have been built previously then cleans them and rebuilds to replace the old build.tools task which failed after deleting the command it was about to run.

added check.tools task before prod build and in the run once for a test build, so any prod/test build will be run with the actual code in /tools even if the existing artifacts are for something else (i.e after you switch branch). 

as far as i could tell it appears if you rebuild the tools before executing any other tasks the rebuilt tasks will be used. my only concern would be any tasks defined in the `gulpfile.ts` probably wouldn't be updated in this case, currently the clean.once task is the only task where this would apply.


mgechev:
Looks good so far! Thanks for the PR! I'll be able to give further feedback a bit later this week.

njs50:
np, no rush. lemme know if you want me to rebase it at some point. i think it covers the basic requirements of:

1) let windows users compile the gulpfile/tools for day to day speed increases in dev builds etc
2) make sure gulpfile/tools are recompiled whenever a build.test or a prod.build is made

if i was on windows i'd probably be tempted to store the git hash of when that stuff was last build and then do a diff vs that hash on each build to see if anything had changed. i'm not sure if rebuilding in a watch would actually work tho, so prob a bit of pissing around to get that working nicely.

tools/tasks/seed/check.tools.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  let checkFile = join(process.cwd(), 'tools', 'config.js');
+
+  // need to require the build.toolchain task as it won't be able to run after we run clear.files
+  let buildTools = require('./build.tools');
+
+  try {
+    fs.accessSync(checkFile, fs.F_OK);
+    util.log('Gulpfile has previously been compiled, rebuilding toolchain');
+    runSequence('clean.tools');
```
mgechev:
```ts
util.log('Running \'build.tools\'');
buildTools(done);
```
Should be run in the callback of `runSequence` since it's asynchronous.

tools/tasks/seed/check.tools.ts
```diff
@@ -0,0 +1,30 @@
+import * as util from 'gulp-util';
+import * as runSequence from 'run-sequence';
+
+import { join } from 'path';
+
+let fs = require('fs');
```
mgechev: Why not `import * as fs from 'fs'` instead to take advantage of type checking?

tools/tasks/seed/build.tools.ts
```diff
@@ -0,0 +1,16 @@
+import * as util from 'gulp-util';
```
mgechev: The naming of this task is much better but let's keep the code from the previous one since it also includes `plugins.template`.

gulpfile.ts
```diff
@@ -152,9 +148,10 @@ let firstRun = true;
 gulp.task('clean.once', (done: any) => {
   if (firstRun) {
     firstRun = false;
-    runSequence('clean.dev', 'clean.coverage', done);
+    runSequence('check.tools', 'clean.dev', 'clean.coverage', done);
```
mgechev: Good catch!

mgechev:
I added some comments. Let me know what you think. The idea with the git hash sounds good! Something like this `git rev-parse upstream/master` should work but we go to some corner cases - what if the developer hasn't named this remote `upstream`, and what if she has cloned from a different branch, etc.

I like the proposed changes they seems like a solid improvement over the current hacky solution for optimization for Windows users.

tools/tasks/seed/check.tools.ts
njs50: that's what i had originally but one of the CI builds failed saying the constants didn't exist there. it did work fine locally tho, and it passed the other CI builds. so maybe there is an issue in a particular node version or something weird?

njs50: fyi if you revert that last commit it's back to using the import. maybe there is something obvious I was doing wrong there?

njs50:
checking the git stuff can be a bit painful. but at least if you only store a hash when the dirs are clean you can then just check against that hash

something like `git diff [hash] --quiet --diff-filter=ACMDR -- ./tools/`

i think is what i use in one of my scripts to see if there are changes in a dir

checking if a dir is clean is the more painful bit. i think i ended up having to do three commands for that to check for staged+uncommitted, uncommitted and untracked files.

tools/tasks/seed/check.tools.ts
njs50: ah right, those shenanigans are because once you run clean.tools you can't run any tasks again until the tools have been rebuilt. now that i reverted to the old build.tools there is a similar issue loading the templateLocals

njs50:
I changed the old build.tools.js to also build the gulpfile, moved the sourcemap write to after the template thing. i had to load the template locals earlier to prevent a require of the compiled ts that no longer exists. 

i think it should be all good now, i rebased/squashed also

tools/tasks/seed/check.tools.ts
mgechev: It is fine, this is a minor thing.

tools/tasks/seed/check.tools.ts
mgechev: I'm not sure I understand what issue you hit here. I'm concerned of potential race condition where you remove files which are built by the current invocation of `build.tools` (in this case this is not going to happen because we have a call `del.sync` but it still seems a bit scary because it's easy to mess this up, it has some indirect coupling).

mgechev:
@njs50 thanks! I left one more comment to keep the discussion about the synchronous run of `clean.tools` and `build.tools`.

tools/tasks/seed/check.tools.ts
njs50: ah i see what you mean now. 'build.tools' can't be in the run sequence as the 'clean.tools' deletes the compiled files for build.tools, so when it attempts to run it fails with the build.tools.js doesn't exist error. if clean.tools was async then we could wait for the promise to resolve before continuing i guess. 

i note your previous commit breaks del, and possibly also rimraf, i'm getting unable to compile typescript errors:

tools/tasks/seed/clear.files.ts (23,22): Argument of type 'string[]' is not assignable to parameter of type 'string'. (2345)

for rimraf and the typings for del doesn't exist anymore

njs50:
the whole having to call other tasks from within the task is a bit nasty, but i'm not sure there is any way around that with the whole deleting the task we are about to run business. that and the not wanting to rebuild the tools unless they were already built kind of forces this interconnection.

anyhow, i've updated the check.tools so that it will not complete until the rebuild is actually complete which should be good for how we use it in the runsequences.

i added a note in the clean.tools that if someone changes it to be async they'd also need to update check.tools to wait.

also fixed the rimraf/del error in clear.files task, tho i'm not sure if that task is actually used for anything?

tools/tasks/seed/check.tools.ts
mgechev: Here are the typings for [rimraf](https://github.com/mgechev/angular-seed/blob/master/package.json#L57).

Would you only change the code to use rimraf and add a comments showing that the task should be kept sync in order to prevent eventual race conditions in future? Otherwise it looks good to me.

mgechev:
@njs50 in general everything looks good to me except the last comment -> `del` to `rimraf` (trying to not introduce new dependencies) and add a comment for the sync tasks to be kept that way and why they are sync.

njs50:
i think rimraf was being used incorrectly in that existing task. after that dependency update the other day it stopped working for me, complaining about not wanting an array of strings to be passed to it, only a single thing to delete.

my understanding was that del is a wrapper for rimraf that lets you run multiple rimrafs at once and returns a promise for all of them.

i could be horribly wrong, but it seemed like of the two del would be the one to keep and rimraf would be the one to drop?

mgechev:
Yes, `del` is only a wrapper around `rimraf` but I don't think we get enough in order to migrate from it. I changed the `clear.files.ts` task to reflect your comment.

njs50:
yeah, probably doesn't make any difference if we're not running things async. if we were running multiple clean tasks at the same time it'd probably make more sense to use del. 

i swapped the del -> rimraf and rebased btw.

mgechev:
Merged. Thanks for the PR!

njs50:
no prob, i like to imagine one day it'll save me some more confusion when a random backend dev accidentally lets their ide compile that stuff. i wonder if the files should be removed from the root tsconfig to prevent that now. that could be bad if the template stuff was doing something important.

PR #1465, opened 2016-10-12
Seed config method to extend system config with other libraries.

troyanskiy:
Ofter we need to extend list of libraries for system config which are in `SYSTEM_CONFIG_DEV` and/or `SYSTEM_BUILDER_CONFIG`.
Method will provide short way to add it.
Can be called from project config constructor
``` ts
this.addPackageBundles({
  name: 'angular2-modal/plugins/bootstrap',
  path: 'node_modules/angular2-modal/bundles/angular2-modal.bootstrap.umd.js',
  packageMeta: {
    main: 'index.js',
    defaultExtension: 'js'
  }
});
```

### DIFF
--- a/tools/config/seed.config.interfaces.ts
+++ b/tools/config/seed.config.interfaces.ts
@@ -11,3 +11,8 @@ export interface Environments {
   [key: string]: string;
 }
 
+export interface ExtendPackages {
+  name: string;
+  path?: string;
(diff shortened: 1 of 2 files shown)


mgechev:
Merged, thanks for the PR!

troyanskiy:
Thanks a lot!

PR #1470, opened 2016-10-13
Sass improvement + examples

lifaon74:
Following: https://github.com/mgechev/angular-seed/issues/1452


mgechev:
Looks like a nice, clean and quite opinionated solution. The Sass support was introduced some time ago but isn't a key feature of the seed.

I'd love to keep the old CSS styles for component and in the same time introduce caching in order to bring better performance for Sass users. Do you think if we revert back to using CSS, using caching in `build.html_css` will bring enough performance improvements? I'm trying to keep the seed as minimalistic as it could be so I'd be happy if we can isolate the Sass specific changes as much as possible.

lifaon74:
I changed css by scss into client/ to provide a demo. It can be ignore safelly (revert to the css version), but we should provide some tutorial or readme, else people won't understand why scss/ folder doesn't compile, etc...

All the changes into tools/ are css safe.

mgechev:
Yes, I agree about the tutorial. We have these options:

- wiki page
- blog page
- fork which supports your config

lifaon74:
Well I'll do a wiki page soon, I'll let you know when it will be done

.sass-lint.yml
```diff
@@ -0,0 +1,18 @@
+rules:
```
mgechev: Linting was dropped in order to reduce the number of dependencies, since it wasn't widely used. Would you mind removing the `.sass-lint.yml` file?

package.json
```diff
@@ -84,6 +84,7 @@
     "gulp-clean-css": "^2.0.11",
     "gulp-concat": "^2.6.0",
     "gulp-concat-css": "^2.3.0",
+    "gulp-debug": "^2.1.2",
```
mgechev: Lets not introduce additional dependencies. Maybe we can report caching using standard `console.log` / `gulp.utils` or just not report it at all?

src/client/app/about/about.component.scss
```diff
@@ -0,0 +1,13 @@
+:host {
```
mgechev: Lets keep it as a CSS file; by using `scss` we implicitly force developers to use Sass.

tools/config/project.config.ts
```diff
@@ -10,6 +10,8 @@ export class ProjectConfig extends SeedConfig {
```
mgechev: Lets keep Sass disabled by default.

tools/tasks/css_task.ts
```diff
@@ -1,9 +1,11 @@
 import { Task } from './task';
+import Config from '../config';
 
 export abstract class CssTask extends Task {
 
+    // Is it still usefull if cache used ?
   shallRun(files: String[]) {
-    return files.some(f =>
+    return Config.ENABLE_SCSS || files.some(f =>
```
mgechev: `ENABLE_SCSS` is not a precondition for execution of this task. We want to run it only if any of the `css`, `scss`, `sass` files have changed.

tools/tasks/seed/build.html_css.ts
```diff
@@ -65,6 +71,22 @@ function processComponentScss() {
 }
 
 /**
+ * Get SCSS Files to process
+ */
+function getSCSSFiles(cacheName:string, filesToCompile:string[], filesToExclude:string[] = []) {
```
mgechev: LGTM

.sass-lint.yml
lifaon74: Hum linting for me is something really important (and should be mandatory into all serious companies...), the .sass-lint.yml provide a decent example and config for scss.

src/client/app/about/about.component.scss
lifaon74: Yep, just here to provide an example

tools/tasks/css_task.ts
lifaon74: Yes, but the piece of code is not properly working because of imports... Sometimes it appends that when I update some assets or sass files, I need to do a new `npm start` else the changes are not detected.

mgechev: I see. Ok, it'll be great if you can add a comment for this. So basically the task will run if the file changed is `.css` or in all cases when we're using Sass, however, we won't have performance implications because of the caching that you introduced. Correct?

.sass-lint.yml
mgechev: Lets keep it as project-specific task, i.e. you can easily introduce it in your project's `tools/project` directory. I noticed that linting in CSS/Sass is rarely used by most developers using the seed so it was a good way to shrink the dev dependencies.

tools/tasks/css_task.ts
lifaon74: Yes it's exact. With the caching, no need to check is any .css or .scss changed.

mgechev:
@lifaon74 looks good! Would you have time to reflect the latest comment so I can merge it?

lifaon74:
So lets recap : 

- I get the last changes from seed,
- I remove examples (sccs files)
- I remove gulp_debug
- I commit build_html, and css_task
- I update the wiki to explain how to use scss files

Right ?

mgechev:
Yes, the listed changes look good to me.

Regarding linting - Sass was not part of the core of the seed project, but it was requestes by different developers so in the end support was introduced in a PR. Although it is nice to have Sass support, because it will be used by, lets say 30% of people, linting in Sass (although being important) is going to be ignored by huge percentage of the Sass users.

In my production project which uses the seed I have css linting enabled, although I dropped it from the seed. Trying to keep the amount of dependencies as small as possible in order to reduce the installation time.

.sass-lint.yml
Bigous: This could be a wiki-howTo

PR #1478, opened 2016-10-14
chore: drop dependencies

mgechev:
Drop 17 npm dev dependencies, css-linting (as not being used often), useless manual type definitions and configuration.


vyakymenko:
Why you drop `compression` ?

PR #1511, opened 2016-10-20
Add staging support as documented

doapp-ryanp:
https://github.com/mgechev/angular-seed#environment-configuration says `staging` is currently supported. This PR adds that.

### DIFF
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -26,6 +26,7 @@ import { Environments, ExtendPackages, InjectableDependency } from './seed.confi
  */
 export const ENVIRONMENTS: Environments = {
   DEVELOPMENT: 'dev',
+  STAGING: 'staging',
   PRODUCTION: 'prod'
 };
 


mgechev:
I'd say that it's better to drop the line from README because there's no `staging.ts` file in `tools/env`.

doapp-ryanp:
How do you feel about adding a `staging.ts` to `tools/env`? Seems like > 80/20 rule for production apps.

To that end, how do you add an env other than `prod` or `dev` since https://github.com/mgechev/angular-seed/blob/master/tools/config/seed.config.ts#L600 only supports those 2?  README.md says its as easy as adding a `<stage>.ts` but this code wont allow that.  Not my intention to come off rude here, just want to make this already awesome project better.

mgechev:
`getEnvironment` gets the build type (environment), not the configuration environment. The configuration environment should be selected [with `--config-env`](https://github.com/mgechev/angular-seed/blob/master/package.json#L9). Probably we should rename this private function to not bring additional confusion.

I'm ok with both options. Maybe it's better to introduce `staging.ts` in `tools/env` because this will help understanding the difference between `--env` and `--config-env` flags.

doapp-ryanp:
Ah gotcha, I'll update this PR with whats discussed here and u can thumbs up or down.

doapp-ryanp:
Ok now looking closer at the code I'm confused.  The [inline docs](https://github.com/mgechev/angular-seed/blob/master/tools/config/seed.config.ts#L60) say `getEnvironment` uses the value of `--config-env`.  [The code](https://github.com/mgechev/angular-seed/blob/master/tools/config/seed.config.ts#L601) however sets the build type by looking for `.prod` in the gulp task name.   My guess is just doc typo but please confirm.

FWIW I'm doing all this work because I have created a gulp task plugin for angular-seed that gzip's assets, adds cache headers (for CloudFront), and invalidates CloudFront cache.  I just want to make sure I understand the angular-seed plumbing here so that I a) get my implementation correct b) make sure its flexible for real world production apps with team(s) of devs

Which gets to my last Q I need answered before I do the re-factor discussed above. 

I want to be able to invoke gulp tasks like `publish.prod` and `publish.staging` and `publish.ryanteammember`.  I want my `publish` task plugin to be able to read a config var out of `project.config.ts` like so:

```js
constructor(){
...
    this.PUBLISH_TASK_CONFIG.s3['prod'] = {
      region: 'us-east-1',
      params: {
        Bucket: 'www.mydomain.org'
      }
    };
    this.PUBLISH_TASK_CONFIG.s3['staging'] = {
      region: 'us-east-1',
      params: {
        Bucket: 'staging.mydomain.org'
      }
    };
    this.PUBLISH_TASK_CONFIG.s3['ryanteammember'] = {
      region: 'us-east-1',
      params: {
        Bucket: 'ryan.mydomain.org'
      }
    };
```

When I call `publish.staging` the "build type" will be `dev` however I want the value of `Config.APP_DEST` to be `dist/prod`.

Questions:

* How do I force a build type of `prod` when my task only contains `publish.staging`?  I could force `Config.ENV` to be `ENVIORNMENTS.PRODUCTION` in my task (ex below) however its too late in the bootstrap process

```
gulp.task('publish.staging', (done: any) => {
  Config.ENV = ENVIRONMENTS.PRODUCTION;
  return runSequence('publish',
    done)
});
```
* When running a Gulp task - how do I get the value of `--config-env`?  I dump the `Config` object from `tools/config` and don't see it defined.  I can hardcode grabbing assets from `dist/prod` dir however, I need a way to lookup the correct index in `Config.PUBLISH_TASK_CONFIG.s3`.  I had been using `Config.ENV` to do this however I now know that wont work per above...

Does this make sense? If not I can put my plugin up on github so you have a full concrete example.

doapp-ryanp:
I got a workaround. Details of my plugin published at [doapp-ryanp/angular-seed-awspublish-task](https://github.com/doapp-ryanp/angular-seed-awspublish-task) 

Let me know if you think the way I set `Config.ENV` in the `gulpfile.ts::gulp.task('awspublish.staging...` will cause un-intended consequences.  thanks.

mgechev:
Yes, the correct way is to use `Config.ENV`. `getEnvironment` got that complicated because we were trying to keep some backwards compatibility. Probably we can simplify it now. I'll open a PR and mention you there in order to take a look.

mgechev:
@doapp-ryanp here's a PR which aims to introduce better semantics of the configuration properties https://github.com/mgechev/angular-seed/pull/1519. There might be still some inconsistencies in the docs but in general I think this brings significant improvement.

Issue #1452, opened 2016-10-10
[Discuss] Sass improvement


lifaon74:
Hi, I'm opening this not to report an issue but to discuss about sass performance.

The problems : 
- last changes (https://github.com/mgechev/angular-seed/commit/20d8d5cbc1f3cc3608de095e951d18f213f9ce64), create sometime problems with sass with import (some files doesn't change but imports yes, problem : they are not re-processed !).
- The scss files are not cached => with medium projects (+30 components -> +30 scss files in app/) it takes about 30s to compute all of them, too long for proper dev.
- Finally, some scss files only contain variables, mixins, etc... and should not be processed, or even added to the main.css when build.prod.

I have done some personnal improvement to the build.html_css task, and share here ideas (maybe not the best, that's what I created a talk). File available [here](http://www.filedropper.com/buildhtmlcss). I'll do a PR if validated.

So first of all : 

1) I created a new folder called scss at the root of `app/APP_NAME`, this folder contains all my "external scss", and none of these files should be compiled.

Into the `css/` folder, I just put one file `main.scss` containing `@import '../scss/main';` This file will import all needed scss files and will be compiled. => This ensure a single entry point, and avoid compiling useless scss files (ex: variables, other projects styles, etc...).

I have also created into project.config.ts a new variable : `SCSS_SRC = `${this.APP_SRC}/scss`;`

2) Into build.html_css

New variable `const useCache = true;` (should go to project.config) to enable/disabled chache.

3 paths : 

``` ts
const appSCSSFiles  = join(Config.APP_SRC, '**', '*.scss'); // all app scss files
const SCSSFiles     = join(Config.SCSS_SRC, '**', '*.scss'); // all "external" scss files that should not be processed
const mainSCSSFiles = join(Config.CSS_SRC, '**', '*.scss'); // entry scss files
```

Then into `processComponentScss()`I used a combinaison of gulp_cache, gulp_progeny, and gulp_filter

``` ts
return gulp.src([
      appSCSSFiles, // all my app scss files
      SCSSFiles // all my "external" scss files
    ])
    .pipe(useCache ? plugins.cached('process-component-scss') : plugins.util.noop()) // check cached files
    .pipe(useCache ? plugins.progeny() : plugins.util.noop()) // check imports
// personal code
    .pipe(debug({title: 'app css before filter'}))
    .pipe(filter([appSCSSFiles, '!' + SCSSFiles])) // remove all "external" scss files that should no be compiled
[...]

PR #1512, opened 2016-10-21
Incorrect path for sourceMappingURL for Angular 2 components

admosity:
Node version: v6.7.0
Chrome Version: 54.0.2840.59 (64-bit)
Chrome Canary: 56.0.2896.3 canary (64-bit)
Firefox Dev Version: 51.0a2

Due to the nature of how Angular 2 handles style injection from components (inject style tag in the head), the sourcemaps are loaded incorrectly for sass (because it's relative to whatever route we're on).

It seems this fix only works on firefox for the time being. Chrome doesn't seem to be able to pick up the sourcemaps still...

Before:
image

pasted image at 2016_10_20 09_17 pm

After:
image

pasted image at 2016_10_20 09_22 pm

+1 #1452 for improving the sass experience!


tools/tasks/seed/build.html_css.ts
```diff
@@ -64,7 +64,12 @@ function processComponentScss() {
     .pipe(plugins.sass(Config.getPluginConfig('gulp-sass')).on('error', plugins.sass.logError))
     .pipe(plugins.postcss(processors))
     .on('error', reportPostCssError)
-    .pipe(plugins.sourcemaps.write(isProd ? '.' : ''))
+    .pipe(plugins.sourcemaps.write(isProd ? '.' : '', {
+      sourceMappingURL: (file: any) => {
+        // write absolute urls to the map files
+        return `/${file.relative}.map`;
```
mgechev: Maybe it'll break if [`APP_BASE` is changed](https://github.com/mgechev/angular-seed/blob/master/tools/config/seed.config.ts#L113). You can use `Config.APP_BASE` instead of `/`.

mgechev:
@admosity would you address this change in order to merge the PR?

admosity:
Yep going to do that now.

admosity:
@mgechev you think `APP_BASE` should be normalized (always add trailing slash)?

mgechev:
No, LGTM.

PR #1519, opened 2016-10-22
refactor: simplify getting build type

mgechev:
A few changes:

- Instead of the `--env` flag for the build type now we use `--build-type`.
- Deprecate the `Config.ENV` property and use `Config.BUILD_TYPE` instead.
- Rename `--config-env` to `--env-config`, although `--config-env` will still work, it is deprecated.
- Change the `env` property of `InjectableDependencies` to `buildType`. Although `env` is still there, the recommended property now is `buildType`.


doapp-ryanp:
I want to use values from `tools/env/staging.ts` in a production `BUILD_TYPE`.  My use case is I have a staging API layer (https://staging.myapi.com) that I want to test my angular app against before doing a `build.prod.exp`, but in general one should be able to do a production build using any app environment config they wish.

As far as I can tell, this is not possible with the current architecture.

I'd like to be able to do something like:

`npm run build.staging.exp  --env-config staging --build-type production`

This would build the app using `tools/env/staging.ts` and spit the output to `dist/prod`.

It is fine to use the gulp task naming convention to branch between `dev` and `prod` `BUILD_TYPE` however I think it must be overridable via configuration.

mgechev:
> I want to use values from `tools/env/staging.ts` in a production `BUILD_TYPE`.

@doapp-ryanp you're able to do that:

```
npm run build.prod.exp --env-config staging --build-type prod
```

Even now you can do that by using

```
npm run build.prod.exp --env-config staging --env prod
```

doapp-ryanp:
Awesome, sorry I didn't gather that by looking at the code.  

If you'd like I can test my https://github.com/doapp-ryanp/angular-seed-awspublish-task plugin against this new architecture.  Let me know if you would like me to test against this `env` branch or if you are going to merge into master right away due to non-breaking function.

mgechev:
@doapp-ryanp sounds awesome if you have time to test your code against the `env` branch. I'm a little bit concerned about the introduced changes, I hope they won't break any (many) projects.

doapp-ryanp:
Few things:

- In the release notes, make sure to tell people they have to modify `src/client/app/main.ts` and `src/client/index.html` to replace `ENV` with `BUILD_TYPE`
- I get lots of `Warning: the "ENV" property is deprecated. Use "BUILD_TYPE" instead.` errors when running `npm run build.prod`. For things like `tools/tasks/seed/css-lint.ts` and `tools/env/prod::ENV`.  Should I be getting warnings here?

mgechev:
Yes, I did this intentionally in order to keep `ENV` and warn people to use `BUILD_TYPE` instead.

doapp-ryanp:
Right, I guess what I'm saying is I'm not using the string `ENV` anywhere in my project (I'm using `env` branch) and I'm still getting that warning msg lots of times with no indication on what code its coming from...

Other than that, the changes in this branch work as intended.  I do see one regression however in the `env` (or maybe `master`) branch, and I can't quite pinpoint where it is coming from.   When running `npm start` I get:

```
http://localhost:5560/node_modules/moment.js
Failed to load resource: the server responded with a status of 404 (Not Found)
localhost/:69 Error: (SystemJS) XHR error (404 Not Found) loading 

http://localhost:5560/node_modules/moment.js(…) "Report this error at https://github.com/mgechev/angular2-seed/issues"
```

For some reason SystemJS is trying to load moment from `node_modules/moment.js` instead of `node_modules/moment/moment.js`.  Here is my relevant app code:

`project.config.ts`:
```javascript
...
    const mtzc = {
      main: 'builds/moment-timezone-with-data-2010-2020.min.js',
      defaultExtension: 'js'
    };

    this.SYSTEM_BUILDER_CONFIG.packages['moment-timezone'] = mtzc;
    this.SYSTEM_CONFIG_DEV.packages['moment-timezone'] = mtzc;
...
```

`manage.component.ts`:
```javascript
...
import * as moment from 'moment-timezone';

@Component({
  moduleId: module.id,
  selector: 'sd-manage',
  templateUrl: 'manage.component.html',
  styleUrls: ['manage.component.css'],
})
...
```

`@types/moment-timezone` does a:
```
...
import * as moment from 'moment';
...
```

My [package.json](https://gist.github.com/doapp-ryanp/674ca6bc6631a3e19ff51e9b795d64ba)

If I run my exact same app against `v2.0.1` release things work fine.  Also puts in `<script src="/node_modules/moment/moment.js?1477346982831"></script>` correctly in `<body>` of `env` branch.

The only way I can get `env` branch to work is by adding the following explicit `moment` loading to the `project.config.ts`.  To be clear I do not have to do this in `v2.0.1`:
```javascript
...
    const m = {
      main: 'moment.js',
      defaultExtension: 'js'
    };

    this.SYSTEM_BUILDER_CONFIG.packages['moment'] = m;
    this.SYSTEM_CONFIG_DEV.packages['moment'] = m;
...
```

mgechev:
Yes, you're getting warnings because `_.template` is accessing this getter. Most likely you have the issue with moment because you need to add `packageConfigPaths` in the systemjs config.

mgechev:
@TheDonDope would you take a look at the PR and share potential breaking changes that you see?

TheDonDope:
Hey @mgechev, i'll have a look at it tomorrow and keep you updated

mgechev:
I did some experiments migrating my internal project to this config. If you have a lot of tasks relaying on the changed configuration properties the migration might be slightly tricker but with keeping the backwards compatibility for a few days we'll let developers know about the change we introduced. I think it's safe.

package.json
```diff
@@ -6,11 +6,11 @@
     "url": "https://github.com/mgechev/angular2-seed"
   },
   "scripts": {
-    "build.dev": "gulp build.dev --color --config-env dev",
+    "build.dev": "gulp build.dev --color --env-config dev --config-env dev",
     "build.dev.watch": "gulp build.dev.watch --color",
     "build.e2e": "gulp build.e2e --color",
-    "build.prod": "gulp build.prod --color --config-env prod",
-    "build.prod.exp": "gulp build.prod.exp --color --config-env prod",
+    "build.prod": "gulp build.prod --color --env-config prod --config-env prod --build-type prod",
```
TheDonDope: `--env-config prod --config-env prod` 

This seems a little confusing. I am about to checkout the PR locally, so this is just my first look at the code. 

We have two flags with a name which only a swap of the two words? :)

TheDonDope: Ohhhhh nevermind, this is the backwards-compatible form of writing it, right?

mgechev: Yes, I will drop these soon.

samuelfernandez:
After updating the upstream with this changes I begun having lots of `Warning: the "ENV" property is deprecated. Use "BUILD_TYPE" instead.`
I've change index.html and main.ts, but I still get it.. Is there anywhere else where I have to replace it? Thanks!

mgechev:
Yes, it's fine to get these warnings. In the next week or so I'll drop the `ENV` property from the config so the warnings will go away.

samuelfernandez:
Ok thanks!

Issue #1550, opened 2016-10-28
Regression: Angular 2.1.2 is breaking AoT build


irsick:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see use [gitter](https://gitter.im/mgechev/angular2-seed) or [stackoverflow](https://stackoverflow.com/questions/tagged/angular2)
```

**Current behavior**
After Angular 2.1.2 update 
npm run build.prod.exp

returns this error
```
[23:30:31] 'compile.ahead.prod' errored after 357 ms
[23:30:31] Error: TSError: ⨯ Unable to compile TypeScript
tools/tasks/seed/compile.ahead.prod.ts (14,10): Supplied parameters do not match any signature of call target. (2346)
    at formatError (/Users/irsick/Documents/angular2-seed/node_modules/gulp/bin/gulp.js:169:10)
    at Gulp.<anonymous> (/Users/irsick/Documents/angular2-seed/node_modules/gulp/bin/gulp.js:195:15)
    at emitOne (events.js:101:20)
    at Gulp.emit (events.js:188:7)
    at Gulp.Orchestrator._emitTaskDone (/Users/irsick/Documents/angular2-seed/node_modules/orchestrator/index.js:264:8)
    at /Users/irsick/Documents/angular2-seed/node_modules/orchestrator/index.js:275:23
    at finish (/Users/irsick/Documents/angular2-seed/node_modules/orchestrator/lib/runTask.js:21:8)
    at module.exports (/Users/irsick/Documents/angular2-seed/node_modules/orchestrator/lib/runTask.js:36:10)
    at Gulp.Orchestrator._runTask (/Users/irsick/Documents/angular2-seed/node_modules/orchestrator/index.js:273:3)
    at Gulp.Orchestrator._runStep (/Users/irsick/Documents/angular2-seed/node_modules/orchestrator/index.js:214:10)
[23:30:31] 'build.prod.exp' errored after 5.92 s
[23:30:31] TSError in plugin 'run-sequence(compile.ahead.prod)'
Message:
    ⨯ Unable to compile TypeScript
tools/tasks/seed/compile.ahead.prod.ts (14,10): Supplied parameters do not match any signature of call target. (2346)
Details:
    diagnostics: [object Object]
Stack:
[...]

PR #1553, opened 2016-10-28
Fixed AoT compilation for angular-cli 2.1.2 (#1550)

irsick:
See https://github.com/mgechev/angular-seed/issues/1550 for details


mgechev:
Hey @irsick, the build fails. Would you take a look at it?

irsick:
@mgechev My change make it work on 2.1.2 but seems it breaks now on previous versions.
Can we update min required Angular version to 2.1.2?

mgechev:
Yes, update to `~2.1.2` will be necessary.

PR #1576, opened 2016-11-01
chore: drop deprecation warnings and deprecated apis

mgechev:


gforceg:
Do these warnings mean the ProdConfig interface needs to be changed to look like this?
export interface EnvConfig {
  API?: string;
  BUILD_TYPE?: string;
}

mgechev:
No, `ENV_CONFIG` is the configuration for the current environment. The build type (i.e. the property `BUILD_TYPE`) is independent from `ENV_CONFIG`. We can have production build with stating environment config, for instance.

TheDonDope:
Hey Minko, do the `build.*` and `serve.*` scripts in the `package.json` also need an update now?
For example: https://github.com/mgechev/angular-seed/blob/master/package.json#L12 ?

mgechev:
Yess, these script params are useless now. I'll drop them in the next commit.

KANekT:
https://github.com/mgechev/angular-seed/blob/master/tools/env/dev.ts - you need update this

mgechev:
No, `ENV` is different from `BUILD_TYPE`.

PR #1593, opened 2016-11-04
(feat) reduced complexity when adding n additional packages

ruffiem:
Thought this was a good idea to prevent piles of `this.addPackageBundles()` in `project.config.ts`


tools/config/project.config.ts
```diff
@@ -31,6 +34,29 @@ export class ProjectConfig extends SeedConfig {
       // {src: `${this.CSS_SRC}/path-to-lib/test-lib.css`, inject: true, vendor: false},
     ];
 
+    // Add packages (e.g. lodash)
+    // this.ADDITIONAL_PACKAGES = [{
```
mgechev: Would you change it to a local variable instead of an instance one? We don't need `ADDITIONAL_PACKAGES` as part of the configuration but only as a container of information which is later intended to be transformed and produce a side-effect.

ruffiem:
@mgechev Weird, I'm checkin out.

PR #1596, opened 2016-11-04
docs(README): add AoT section

Nightapes:


Nightapes:
@TheDonDope i added it to the table of contents

PR #1603, opened 2016-11-06
Specify ENABLE_SCSS setting via environment variable

hank-ehly:
Some npm run-scripts contain multiple gulp comman-
ds (ex. e2e.ci). Multiple gulp commands may refer-
ence the ENABLE_SCSS setting; however, '--scss' is
only appended to the final command:

    gulp ... && gulp ... && gulp ... "--scss"

This results in a build error.

To provide each command in an npm run-script the
correct ENABLE_SCSS setting, set the ENABLE_SCSS
setting using an environment variable rather than
a command-line argument.

BREAKING CHANGES
1. The '--scss' no longer has any effect
2. To enable SASS support via the command line,
   one must prefix their run-script command with
   ENABLE_SCSS=1 or ENABLE_SCSS=true.

### DIFF
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -282,11 +282,11 @@ export class SeedConfig {
   VERSION_NODE = '4.0.0';
 
   /**
-   * The flag to enable handling of SCSS files
-   * The default value is false. Override with the '--scss' flag.
+   * Enable SCSS stylesheet compilation.
+   * Set ENABLE_SCSS environment variable to 'true' or '1'
    * @type {boolean}
    */
-  ENABLE_SCSS = argv['scss'] || false;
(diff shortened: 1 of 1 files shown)


hank-ehly:
May I please see what went wrong with the appveyor build?

mgechev:
Maybe we can keep both for backwards compatibility?

```
ENABLE_SCSS = ['true', '1'].indexOf(`${process.env.ENABLE_SCSS}`.toLowerCase()) !== -1 || argv['scss'] || false
```

hank-ehly:
@mgechev 

Great idea. I've added back in the `--scss` flag.
It should be noted that use of the `--scss` flag with run-scripts including the AND operator could result in an error. I will update the necessary documentation with this information if this pull request is merged to master.

PR #1609, opened 2016-11-06
chore: add vscode task for tslint

robstoll:
Might be helpful for others as well. Running tslint via vscode task shows all the errors in the Problems pane


.vscode/tasks.json
```diff
@@ -0,0 +1,30 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
```
mgechev: Let's drop this comment. It's not an official JSON construct and it created problems in the past (in `tslint.json`).

PR #1624, opened 2016-11-11
chore: add vscode task for tslint

robstoll:
Might be helpful for others as well. Running tslint via vscode task shows all the errors in the Problems pane

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,28 @@
+{
+  "version": "0.1.0",
+  "command": "npm",
+  "isShellCommand": true,
+  "showOutput": "always",
+  "suppressTaskName": true,
+  "tasks": [
+    {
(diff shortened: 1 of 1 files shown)


mgechev:
I am holding this because most developers are using the `tslint` plugin and I don't see why we should force them to use other way of reporting the tslint errors but on the other hand, this is the way codelyzer will lint external templates.

In a few days I'll review it again and merge, thanks for addressing my comment!

robstoll:
I see it more as a supplement. I am using the `tslint` plugin myself but it is configured to lint in the open file. If I want to lint all sources I was running`npm run lint` which showed the errors on the console. That's why I created the task, so I see the errors in the Problems pane and can jump to the problems. Just as information why I even have this task. I am using it very seldom (mainly when I change tslint.json), thus I am perfectly fine if it is not merged at all.

PR #1627, opened 2016-11-13
routing corresponding to STYLE 02-12

robstoll:
- removed .routes.ts files and replaced them with -routing.module.ts files (corresponding to [Angular Style Guide 02-12](https://angular.io/styleguide#02-12))
- removed extra blank line between Decoration and class

### DIFF
--- a/src/client/app/app-routing.module.ts
+++ b/src/client/app/app-routing.module.ts
@@ -1,0 +1,15 @@
+import { NgModule } from '@angular/core';
+import { RouterModule } from '@angular/router';
+
+@NgModule({
+  imports: [
+    RouterModule.forRoot([
+      /* define app module routes here, e.g., to lazily load a module
+         (do not place feature module routes here, use an own -routing.module.ts in the feature instead)
+       */
+    ])
+  ],
+  exports: [RouterModule]
+})
+export class AppRoutingModule { }
+
--- a/src/client/app/home/home-routing.module.ts
+++ b/src/client/app/home/home-routing.module.ts
@@ -1,0 +1,13 @@
+import { NgModule } from '@angular/core';
+import { RouterModule } from '@angular/router';
+import { HomeComponent } from './home.component';
+
+@NgModule({
+  imports: [
+    RouterModule.forChild([
+      { path: '', component: HomeComponent }
+    ])
+  ],
+  exports: [RouterModule]
+})
+export class HomeRoutingModule { }
--- a/src/client/app/about/about-routing.module.ts
+++ b/src/client/app/about/about-routing.module.ts
@@ -1,0 +1,13 @@
+import { NgModule } from '@angular/core';
+import { RouterModule } from '@angular/router';
+import { AboutComponent } from './about.component';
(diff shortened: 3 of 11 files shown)


netstart:
What's the benefit of doing it this way?

TheDonDope:
@netstart To keep up the consistency with the styleguide
@robstoll Thanks for the PR! I'll take a closer look on it later today

src/client/app/home/home-routing.module.ts
```diff
@@ -0,0 +1,13 @@
+import { NgModule } from '@angular/core';
+import { RouterModule } from '@angular/router';
+import { HomeComponent } from './home.component';
+
+@NgModule({
+  imports: [
+    RouterModule.forChild([
+      { path: 'about', component: HomeComponent }
```
TheDonDope: Shouldn't the path be `'home'` ?

src/client/app/app-routing.module.ts
```diff
@@ -0,0 +1,13 @@
+import { NgModule } from '@angular/core';
+import { RouterModule } from '@angular/router';
+
+@NgModule({
+  imports: [
+    RouterModule.forRoot([
+      /* define app non-feature routes here */
```
TheDonDope: More of a general question: Since we both have the `HomeModule` and `AboutModule`, which routes would be considered 'non-feature' routes? Sure, at the moment we don't have any. But in a broader context, what routes would be defined in the `AppRoutingModule` ?

src/client/app/home/home-routing.module.ts
robstoll: sorry about that. I wanted to see how angular deals with routing conflicts in Feature Modules (seems like first wins) and I accidentally did not switch it back. I'll create a fix

src/client/app/app-routing.module.ts
robstoll: I am fairly new to angular and I figured it is a good practice to switch the routing to the new style guide. [In some places](https://angular.io/docs/ts/latest/guide/router.html#!#the-heroes-app-code) in the doc the app-routing.module.ts is more or less the same as the one here -> empty. In another place it is used for [lazy loading modules](https://angular.io/docs/ts/latest/guide/router.html#!#lazy-loading-route-configuration).

robstoll:
Ah.. path should be '.' I'll commit a fix later

src/client/app/app-routing.module.ts
TheDonDope: Ah thanks for the doc links! Hehe, yeah specifically with the 'new' router it is often kind of hard to find the most up-to-date info :)

robstoll:
@TheDonDope Is there still something open?

TheDonDope:
@robstoll nah its fine, but you need to rebase onto current master, since your branch is behind current master.

robstoll:
I've rebased it again, do not wait too long, there are too many contributors supplying pull requests :)

mgechev:
LGTM

PR #1631, opened 2016-11-15
- Fix resolution of external dependencies during production build on Linux

ZuSe:
This resolves https://github.com/mgechev/angular-seed/issues/1619


mgechev:
Thanks for resolving the issue and opening the PR! Sorry I didn't have chance to take a look at the problem closer.

PR #1636, opened 2016-11-16
Removed @angular/tsc-wrapped as it is a peer dependency

jesperronn:
Also updated yarn lockfile, which was not done.

According to tsc-wrapped page, it's an internal dependency

> tsc-wrapped
>
> This package is an internal dependency used by @angular/compiler-cli. Please use that instead.

See https://github.com/angular/angular/tree/master/tools/@angular/tsc-wrapped


jesperronn:
Thanks!

mgechev:
Thanks for the PR!

PR #1638, opened 2016-11-17
Dev dependency

netstart:
- move intl and minimath to dev dependency
- add instructions to generate api documentation in readme


mgechev:
intl is a polyfill used in the dev and prod bundles, so it needs to be part of the dependencies object. Regarding minimatch you're correct, it is incorrectly placed in dependencies and should be in the devDependencies.

PR #1642, opened 2016-11-18
Implement async/await for e2e tests

VladimirMakaev:
This is originated in advanced-seed project 
https://github.com/NathanWalker/angular-seed-advanced/pull/306

Async/Await is only supported in es2015+ target (and it doesn't seem to be supported for es5 anytime soon) but that's a very nice and neat feature for writing e2e tests as many constructs in nodejs selenium is based on promises and as tests are getting more complicated more and more friction is expected when using promises.

In order to do that e2e tests should be built with another tsconfig.json file. But we still want to compile the main application for es5

This PR consist of the following (slight changes from the angular-seed-advanced project)

- e2e tests and framework moved outside of client folder into e2e folder. This is required to support proper detection of another project by VS Code. Separate ```tsconfig.json``` is used for that folder
- additional config variables added to E2E locations to follow the seed patterns
- typing is added to ```makeTsProject``` to make it nicer to use
- the option ```{ target: 'es2015' }``` is used in ```build.js.e2e.ts``` is used as a workaround to the global variable used inside of the ```makeTsProject``` to support more than one project at the same time. (We could think of fixing it in some way but I don't think a new function is a good idea either)
- ```tslint``` is running against e2e files too
- ```tslint``` is executed with ```serve.e2e```

### DIFF
--- a/protractor.conf.js
+++ b/protractor.conf.js
@@ -2,7 +2,7 @@ const config = {
   baseUrl: 'http://localhost:5555/',
 
   specs: [
-    './dist/dev/**/*.e2e-spec.js'
+    './dist/e2e/**/*.e2e-spec.js'
   ],
 
   exclude: [],
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -31,11 +31,9 @@ gulp.task('build.dev.watch', (done: any) =>
 // --------------
 // Build e2e.
 gulp.task('build.e2e', (done: any) =>
-  runSequence('clean.dev',
+  runSequence('clean.e2e',
               'tslint',
-              'build.assets.dev',
               'build.js.e2e',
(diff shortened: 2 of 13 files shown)


mgechev:
@VladimirMakaev looks awesome! I'll take a further look into the changes during the weekend and merge/comment as soon as possible! Thank you!

protractor.conf.js
```diff
@@ -2,7 +2,7 @@ const config = {
   baseUrl: 'http://localhost:5555/',
 
   specs: [
-    './dist/dev/**/*.e2e-spec.js'
+    './dist/e2e/**/*.e2e-spec.js'
```
mgechev: Can you use the path from `Config` here?

PR #1646, opened 2016-11-19
Simplify proxy middleware configuration

karlhaas:
This change allows to add several proxy middleware configurations by adding them in project.config.ts only.
Without this change we had to add the middlewares in project.config.ts, e2e.ts and server.ts

### DIFF
--- a/tools/tasks/seed/e2e.ts
+++ b/tools/tasks/seed/e2e.ts
@@ -1,13 +1,17 @@
 import * as express from 'express';
 import * as history from 'express-history-api-fallback';
 import * as gulp from 'gulp';
-import { resolve } from 'path';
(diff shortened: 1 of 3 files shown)


tools/tasks/seed/e2e.ts
```diff
@@ -22,11 +26,12 @@ class Protractor {
  * Executes the build process, running all e2e specs using `protractor`.
  */
 export = (done: any) => {
+  process.env.LANG='en_US.UTF-8';
   new Protractor()
-    .server(5555, './dist/prod')
+    .server(Config.PORT, Config.PROD_DEST)
```
mgechev: Good catches!

Issue #645, opened 2016-03-17
Improve (speed up) build.dev.watch task


mpetkov:
First off I wanted to say how much I love this seeder.  It is very easy to use and follow.  One possible improvement that I could suggest would speed up the build.dev.watch task.  Currently, no matter what file changes, the entire dev destination gets rebuilt from scratch.  So that means that, on average, each change takes about 10-15 seconds to complete.  It would be faster if the watch was separated out so that it only updates files that were changes, instead of the whole directory.  So if I change something on a css/sass file, only that file should compile and the others (such as build.js.dev) should not.  This will make previewing in dev environment faster and less painful.

d3viant0ne:
The change after the initial watch takes ~2.5 seconds, there was another discussion around this #496

The typescript build is incremental, an incremental build for things like css probably wouldn't be viable if you think about preprocessors ( which are supported ).

Rebuilding everything is the simple solution, rebuilding what changed is certainly possible but I'd have to tinker with it  to see if it could be done in a way which can be generic enough for the seed tools.

Perhaps a possible feature after the cli project is finished.

//cc @mgechev @ludohenin

mpetkov:
Thank you for the answer.  Looks like from your comment in #496 node V5 is way faster than V4 (which I am using).  In addition, when I have time I will try to separate the watchers so that they only happen per type so that if a css/html/whatever change happens, the typescript rebuild (which is the most time consuming) doesn't trigger.

mgechev:
@mpetkov it will be great if you can help with this! Let me know what you find for node 5 vs 4, for your project.

d3viant0ne:
@mgechev - Node 5 is definitely faster than 4. 

- Mid 2014 ( 16gb + SSD )
- Fresh checkout of the angular2-seed

**Node 4.4.0**

**Average of 5 runs dropping highest and lowest**
Initial build.js.dev - 6.5s 
Watch build.js.dev - 4.1s

**Node 5.9.0**

**Average of 5 runs dropping highest and lowest**
Initial build.js.dev - 4.1s 
Watch build.js.dev - 1.8s

mpetkov:
@mgechev - Unfortunately my work hasn't approved Node 5 so I am unable to install it on my work machine.  Here are my numbers on the latest version of node I am able to install:

Node 4.3.2:
Machine Specs: Windows 7; 8GB Ram; 2.6ghz i7-5600U
[...]

PR #1649, opened 2016-11-21
Re-enabled css streaming #645

totev:
This mini-fix re-adds the css streaming support from browsersync to the seed.


mgechev:
Not sure if this is a good idea because changes in component styles will not be applied. External stylea dont have to be injected but compiled into the component's context instead.

PR #1655, opened 2016-11-22
Fix nginx template file name in docker-compose.production.yml

miltador:
Just a small fix in docker configuration for production. The file name of template was wrong so the docker could not start properly.

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -26,7 +26,7 @@ services:
     ports:
       - '5555:80'
     volumes:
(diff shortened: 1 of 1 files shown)


TheDonDope:
Ooops! Thanks for the catch! I copy-pasted the config from another project of mine :)

TheDonDope:
@miltador Can you please add another commit with `$ git commit --amend` and `$ git push origin miltador-fix-docker-prod --force` to trigger a new Patchset to be uploaded here to this PR? The Travis CI build failed, and we need to trigger it again. Thanks!

miltador:
@TheDonDope sure, done! 🚀

PR #1656, opened 2016-11-22
Update wiki page contributions list

amalshehu:
Add material2 integration instructions.


TheDonDope:
I guess we don't need a manual update on this. I think @mgechev has a script which updates the wiki contributors. So the list gets updated from time to time.

amalshehu:
@TheDonDope Thanks for the information friend. It was a mistake.

TheDonDope:
@amalshehu No worries :)

PR #1659, opened 2016-11-23
Fix failing AoT build

hank-ehly:
As of angular/angular [b15039d](https://github.com/angular/angular/blob/b15039d228d206c8c460eca81af51a78f2fec12e/modules/%40angular/compiler-cli/src/codegen.ts) the 'transitiveModules' option has
been removed from compiler-cli CodeGenerator's 'codegen' method.
The current seed still includes this option, which causes the production AoT build to fail. 
Removing the option -- just like has been done in the angular-cli source code -- allows the production AoT build to execute successfully.

### DIFF
--- a/tools/tasks/seed/compile.ahead.prod.ts
+++ b/tools/tasks/seed/compile.ahead.prod.ts
@@ -10,7 +10,7 @@ import Config from '../../config';
 function codegen(
   ngOptions: AngularCompilerOptions, cliOptions: NgcCliOptions, program: ts.Program,
   host: ts.CompilerHost) {
-  return CodeGenerator.create(ngOptions, cliOptions, program, host).codegen({ transitiveModules: true });
+  return CodeGenerator.create(ngOptions, cliOptions, program, host).codegen();
 }
 
 const modifyFile = (path: string, mod: any = (f: string) => f) => {


hank-ehly:
It looks like 'yarn' is causing the travis build to fail.

mgechev:
Thanks for the PR!

koodikindral:
It causes soome backward-compatibility issues as this parameter is still present in codegen.d.ts. Please laso see #1666 
```
...
export declare class CodeGenerator {
    private options;
    private program;
    host: ts.CompilerHost;
    private staticReflector;
    private compiler;
    private reflectorHost;
    constructor(options: AngularCompilerOptions, program: ts.Program, host: ts.CompilerHost, staticReflector: StaticReflector, compiler: compiler.OfflineCompiler, reflectorHost: StaticReflectorHost);
    private calculateEmitPath(filePath);
    codegen(options: {
        transitiveModules: boolean;
    }): Promise<any>;
    static create(options: AngularCompilerOptions, cliOptions: NgcCliOptions, program: ts.Program, compilerHost: ts.CompilerHost, reflectorHostContext?: ReflectorHostContext, resourceLoader?: compiler.ResourceLoader, reflectorHost?: ReflectorHost): CodeGenerator;
}
...
```

PR #1675, opened 2016-11-26
feat(tasks): allow easier overriding of existing tasks

### LINKED ISSUE #1504: [idea] Make customizing tasks without editing gulpfile easier
This is an idea I think would help the community if implemented. Pls let me know what you think

**the challenge**
We need to customize tasks without editing `gulpfile.ts` since that will be replaced (and our edits lost) when the repository is updated

**the current solution**
We can overwrite any task in `tools/tasks/seed/` by writing our own, giving it the same name as the built-in task, saving it in `tools/tasks/project/`. So far so good

**what's still missing**
Sometimes we need to change the tasks defined in `gulpfile.ts` itself, and all we need to do is one small change that doesn't require a full rewrite. For instance here is my use case: once the project is built and files are placed in destination folder, I need to move them all into a remote folder from which my webserver actually serves the files. To accomplish this right now, I have two options:
1. Copy every build task  in `gulpfile.ts` and paste it in `tools/tasks/project/` as one of my own that does the exact same thing as the built-in, but has one additional step.
2. Edit `gulpfile.ts` to insert my step into every build task. For instance:

```
// Build dev.
gulp.task('build.dev', (done: any) =>
  runSequence(
              ...,
              'push', // I inserted my task to move built files to a remote folder
              done));
// --------------
// Build prod.
gulp.task('build.prod', (done: any) =>
  runSequence(
              ...,
              'push', // I inserted my task to move built files to a remote folder
              done));
```

Neither of these options is very good.

**my proposal**
For the built-in `build` and `serve` tasks in `gulpfile.ts`, add a step at the beginning and a step at the end, that will serve as hooks for the developer to customize the task.  For example:

```
// Build dev.
gulp.task('build.dev', (done: any) =>
  runSequence(
              'pre.build.dev'
              ...,
              'post.build.dev'
              done));
```

The files `tools/tasks/seed/pre.build.dev` and `tools/tasks/seed/post.build.dev` would do nothing. However this gives the developer the opportunity to create files by the same name in `tools/tasks/project` that would do something useful.

For me, I could perform my `push` tasks within `tools/tasks/project/post.build.dev`, without the need to edit `gulpfile.ts`.

PS: similar hooks could be inserted into other `build` and `serve` tasks. What do you think?

mgechev:
Fix #1504

// cc @ludohenin @TheDonDope @NathanWalker


README.md
```diff
@@ -340,8 +340,9 @@ Forks of this project demonstrate how to extend and integrate with other librari
 │   │   ├── banner-256.txt
 │   │   ├── banner.txt
 │   │   ├── project.config.ts  <- configuration of the specific project
+│   │   ├── project.tasks.ts   <- override composite gulp tasks
 │   │   ├── seed.config.interfaces.ts
-│   │   └── seed.config.ts     <- generic configuration of the seed project
+│   │   ├── seed.tasks.ts      <- default composite gulp tasks
```
robstoll: I guess you did not delete the description about seed.config.ts on purpose

PR #1680, opened 2016-11-27
example for env-config usage in index.html

robstoll:
I figured an example might be helpful for beginners.


mgechev:
Yes, not bad. Maybe we can be a bit more artistic with the favicon?

robstoll:
Absolutely, I could not find an official icon that's why I have chosen such simplistic favicons. Please replace them with better looking versions

mgechev:
@robstoll if you have an idea for something fancier you're welcome to do the change yourself. I'm not a good designer myself so I'm always open for suggestions :-).

PR #1691, opened 2016-12-04
tslint should not lint d.ts files

robstoll:
- thus added a `-e` option to the tslint command
- fixed the tslint task for vscode
- and fixed a tslint issue "missing whitespace"


mgechev:
Thanks for the improvements!

PR #1694, opened 2016-12-06
Add gulp-tslint back in. It now supports tslint v4

nmelv170:
I simply added the latest version of gulp-tslint back into package.json and reverted the gulp task for tslint back to its former state. npm install and test both run and pass. tslint task was able to find issues when manually introduced into the code.

PR where dependency was removed: https://github.com/mgechev/angular-seed/pull/1681

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "gulp-sourcemaps": "2.2.0",
     "gulp-template": "^4.0.0",
     "gulp-typescript": "^3.0.2",
+    "gulp-tslint": "7.0.1",
     "gulp-uglify": "^2.0.0",
     "gulp-util": "^3.0.7",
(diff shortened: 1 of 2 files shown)


TheDonDope:
Hi @nmelv170, do these changes from here need to be reverted too? https://github.com/mgechev/angular-seed/commit/57c655a1a81e29a307bd6b01329d2ad9153fe25b

nmelv170:
@TheDonDope, I don't think so. Either way the linting will be done. this just gets it back into the gulp builds where the task was disabled.

mgechev:
@nmelv170 seems right that the `package.json` script need to be updated. I'll fix this in a separate PR.

mgechev:
Fixed `package.json`. `.vscode/tasks.json` seems fine.

PR #1709, opened 2016-12-10
[WIP] chore: update deps

mgechev:


SamVerschueren:
I have issues with AoT if I just update the deps like this. I believe some things should be changed in the `compile.ahead.prod.ts`. I'll let you know if I found out what exactly.

SamVerschueren:
It was failing for me because I used `typescript@2.1.x` instead of `typescript@2.0.x`. So some things broke between these two minor versions.

Issue #1736, opened 2016-12-26
e2e.ts file


csreddy:
Where is e2e.ts used ? Definitely not for e2e task as the code invoked for e2e task does not seem to rely on this file ? Should this be removed ?

Also there is some discrepancy with code
```javascript
export = (done: any) => {
  new Protractor()
    .server(5555, './dist/prod')   <--- refers to PROD build
    .then((server: any) => {
      gulp
        .src('./dist/dev/**/*.e2e-spec.js') <----- refers to DEV build
        .pipe(protractor({ configFile: 'protractor.conf.js' }))
        .on('error', (error: string) => { throw error; })
        .on('end', () => { server.close(done); });
    });
};
```

mgechev:
Yes, it seems `e2e.ts` is not in use anymore and should be removed. We need to revise all tasks and drop the unused ones.

If you have some time a PR will be appreciated!

mgechev:
@csreddy `gulp e2e` is used for running e2e tests in CI.

PR #1739, opened 2016-12-29
#1736: Delete the e2e.ts file since it is no longer used

mithun-daa:
#1736 Fix


mgechev:
@mithun-daa thanks for the PR! It seems `gulp e2e` is used in CI, excuse me for the mistake...

PR #1768, opened 2017-01-15
feat: User defined watch paths

admosity:
Allow users to use project settings to include external directories aside from app src directory to trigger compilation.

Example config from actual project:

image

Useful for when coupled with gulp-sass includePaths settings:
image


mgechev:
Thanks for the PR!

Issue #1670, opened 2016-11-24
WebStorm (v2.3.1): Live reload after file change (html) does not work. SyntaxError: Unexpected token < in JSON at position 0


webia1:
possible bug ?
Windows 7, npm 4.0.1, node v6.8.1, TypeScript 2.0.10 

**Initial seed, nothing has been changed.** [Work-around here](https://github.com/mgechev/angular-seed/wiki/Known-Issues) does not work.

That's the Chrome Console: 

![webstorm_2 3 1_live_reload](https://cloud.githubusercontent.com/assets/2741558/20605988/7473fe88-b26e-11e6-9c4e-8db105559d8e.png)

``` javascript
VM661:1 Uncaught SyntaxError: Unexpected token < in JSON at position 0
    at JSON.parse (<anonymous>)
    at http://localhost:5555/app/system-config.js:1:20(anonymous function) @ system-config.ts:2

    System.config(JSON.parse('<%= SYSTEM_CONFIG_DEV %>'));
    //# sourceMappingURL=system-config.js.map

(index):48 (SystemJS) Unexpected token <
	SyntaxError: Unexpected token <
	    at eval (<anonymous>)
	Evaluating http://localhost:5555/app/main
	Error loading http://localhost:5555/app/main Not expecting this error? Report it at https://github.com/mgechev/angular-seed/issues
```

As already [mentioned here](https://github.com/mgechev/angular-seed/wiki/Known-Issues) safe-write is activated:

![webstorm_safe_write](https://cloud.githubusercontent.com/assets/2741558/20606010/99ca30d0-b26e-11e6-9b2e-0ea406bf96dd.png)

mgechev:
I don't have webstorm installed, neither a license so it'll be hard to reproduce...

webia1:
You can always download an early access version (30 Days) here: https://confluence.jetbrains.com/display/WI/WebStorm+EAP

At reloading-time the `'<%= SYSTEM_CONFIG_DEV %>'` cannot be interpreted, whatever the problem is, it is time or order-related.

TheDonDope:
Sounds a hell lot like https://github.com/mgechev/angular-seed/issues/1625#issuecomment-260276901

webia1:
I've asked the question at stackoverflow: http://stackoverflow.com/questions/40805944/webstorm-v2-3-1-live-reload-after-file-change-html-does-not-work-syntaxerr?noredirect=1#comment68832698_40805944

[...]

PR #1773, opened 2017-01-19
Fix issue with html only changes breaking output

traviscollins:
This one line change fixes the following issue:

1) Checkout master branch
2) npm install
3) npm start
4) Wait for browser to open
5) Add just one space to any app/*.html file and save
6) Browser refreshes and fails to load app

Origin of fix and explanation here:
https://github.com/NathanWalker/angular-seed-advanced/issues/318#issuecomment-265957882

### DIFF
--- a/tools/tasks/seed/build.assets.dev.ts
+++ b/tools/tasks/seed/build.assets.dev.ts
@@ -14,6 +14,7 @@ export =
       let paths: string[] = [
         join(Config.APP_SRC, '**'),
         '!' + join(Config.APP_SRC, '**', '*.ts'),
+        '!' + join(Config.APP_SRC, '**', '*.js'),
         '!' + join(Config.APP_SRC, '**', '*.scss'),
         '!' + join(Config.APP_SRC, '**', '*.sass')
             ].concat(Config.TEMP_FILES.map((p) => { return '!' + p; }));


mgechev:
Thanks for the PR!

Here's an issue related to your fix https://github.com/mgechev/angular-seed/issues/1670. What bothers me a bit in the fix is that it may break existing projects which contains mixture between TypeScript and JavaScript. Maybe we can merge it and wait a bit to see what feedback we're going to get from people.

traviscollins:
Definitely related to #1670. 

For further info - I use IntelliJ in the latest version of OSX. 

I don't full understand the fix in the pull request (how js files ended up in the /src folder because of IntelliJ) - but figured I'd submit it and see what you thought about it.

mgechev:
Thanks for the help!

ultimatedion:
I may be wrong about my understanding of the fix... 
but while it may work, I don't believe that this is the right fix for the issue. It may cause other problem.

It takes an « unwanted file »(a file that was **not** generated by the build but generated by intelliJ) through the build process. Of course, the token <%= SYSTEM_CONFIG_DEV %> gets replaced by the proper value and it looks like it fixes the problem.

One problem I am thinking of with this fix:
1- IntelliJ compiles your .ts to .js in the src folder.
2- You turn off the IntelliJ compiler
3- You do not clear the .js located in src folder generated by IntelliJ compiler
4- You make a modification to a .ts file.
5- I don't know the build order but the leftover .js files previously generated by IntelliJ might override the .js files generated(from your .ts files) by the build process.

I have a question in mind.. why a change to an HTML file triggers the build/copy of .ts or .js files to dist? I think that theses files shouldn't be updated in the dist folder. This would probably fix the issue properly.

PR #1784, opened 2017-01-24
Add GIT_DESCRIBE to Config and task git.describe in order to be able …

karlhaas:
…to get a version information for debugging

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "gulp-concat": "^2.6.0",
     "gulp-concat-css": "^2.3.0",
     "gulp-filter": "^4.0.0",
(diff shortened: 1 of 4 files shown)


mgechev:
Lets not introduce additional dependencies. If someone requests this feature we can introduce it.

karlhaas:
Do you mind adding initialize to all tasks? In seed.tasks it can be empty but it would make it easier to add custom initialize tasks to the pipeline.
Without it I have to move all tasks to project.tasks which complicates updates.
I would reduce the pull request to just this change if you agree ...

mgechev:
The initialization task is merged. Thanks!

PR #1786, opened 2017-01-25
Remove obsolete types from prod rollup aot ts config and and add  rol…

karlhaas:
…lup-plugin-includepaths

rollup-plugin-includepaths is needed when tsconfig paths mappings are used within the project.


brian428:
I'm hitting an issue with the rollup build that maybe this will fix. I get a build error saying:
```
TSError: ⨯ Unable to compile TypeScript
tools\tasks\seed\build.js.prod.rollup.aot.ts (19,5): Argument of type '{ target: "es2015"; module: "es2015"; moduleResolution: string; types: string[]; }' is not assignable to parameter of type 'Settings'.
```

Looking at the gulp-typescript Settings interface (https://github.com/ivogabe/gulp-typescript/blob/master/release/main.d.ts), there is no `types` property, hence this error. In fact, I'm not sure how the current rollup build could ever have worked, given this type mismatch?

karlhaas:
@brian428 I'm not sure but I think that upgrading gulp-typescript to version 3.1.4 solved this issue

tools/tasks/seed/remove.debug.ts
```diff
@@ -0,0 +1,13 @@
+import * as gulp from 'gulp';
+import { join } from 'path';
+import Config from '../../config';
+const removeDebug = require('gulp-remove-logging');
+
+/**
+ * Remove all debug statements from distribution
+ */
+export = () => {
```
mgechev: Uglifyjs removes `debugger` statements. I'm not sure if we should remove `console.log` statements by default.

mgechev:
@brian428 were you able to resolve the problem?

karlhaas:
@mgechev now I use ugilfyjs to remove debug statements and it must be enabled in project config

brian428:
@mgechev I had to change the typing for the TS project config object (e.g. the `makeTsProject()` in `tsproject.ts`) to use `any` instead of `ts.Settings`. Basically, the `Settings` interface in gulp-typescript appears to be pretty far out of date. However, I think gulp-typescript really just forwards the config into tsc, so using `any` allows properties that gulp-typescript doesn't have in it's `Settings` interface to be passed through.

mgechev:
@karlhaas thanks for the PR! I'd prefer to not merge it at this point and reconsider it once the implemented features are requested by people using the seed. I'm afraid of not constantly expanding the scope of the project until we reach a point when it's complexity get unmanageable.

karlhaas:
@mgechev maybe I put to many changes in one PR.

- The changes in build.js.prod.rollup.aot.ts are improvements and should make it more consistent with other tasks. The current version does not call _.template and by overwriting types they have to be maintained in several files.

- Drop console is a feature which can be removed. In the current PR it is just an optional parameter and should not increase the complexity (it should not be in project.config.ts)

- The changes in build.bundles.app.rollup.aot.ts are needed if someone uses the baseUrl and paths feature of Typescript 2. Without the change rollup does not find the imports. Without commonjs.namedExports immutable.js does not work.

- The changes in copy.prod.rollup.aot.ts are needed if someone wants to add moment.js. Maybe there is a better solution for this ...

Please verify the changes and let me know which of them I should remove. I'm sure that some changes improve the seed without adding complexity. Thx.

tools/config/project.config.ts
```diff
@@ -43,6 +43,8 @@ export class ProjectConfig extends SeedConfig {
 
     /* Add to or override NPM module configurations: */
     // this.mergeObject(this.PLUGIN_CONFIGS['browser-sync'], { ghostMode: false });
+
+    this.PROD_DROP_CONSOLE = true;
```
mgechev: Lets remove this.

tools/config/seed.config.ts
```diff
@@ -307,6 +307,12 @@ export class SeedConfig {
   EXTRA_WATCH_PATHS: string[] = [];
 
   /**
+   * Set to true in order to remove all console statements from your source
+   * @type {boolean}
+   */
+  PROD_DROP_CONSOLE: boolean = false;
```
mgechev: Lets remove this.

tools/tasks/seed/minify.bundles.ts
```diff
@@ -10,7 +10,9 @@ const plugins = <any>gulpLoadPlugins();
 const getTask = (target: string, destDir: string) => {
   return gulp.src(join(destDir, target))
     .pipe(plugins.uglify({
-      compress: true,
+      compress: {
```
mgechev: Lets remove this.

mgechev:
@karlhaas the changes make sense. Hopefully we won't hit issues with `gulp-typescript` version mismatch. Lets drop the logic for removal of log statements.

PR #1809, opened 2017-02-10
Fix: issue of not overridable PORT or any other property in ProjectConfig

alllx:
Fix: issue of not overridable PORT or any other property which is used in other property objects of SeedConfig with ProjectConfig

 - renamed CSS_PROD_BUNDLE to CSS_BUNDLE_NAME;
 - changed method of adding proxy middleware;

### DIFF
--- a/tools/config/project.config.ts
+++ b/tools/config/project.config.ts
@@ -27,7 +27,6 @@ export class ProjectConfig extends SeedConfig {
 
     // Add `local` third-party libraries to be injected/bundled.
     this.APP_ASSETS = [
-      ...this.APP_ASSETS,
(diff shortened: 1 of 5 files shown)


mike42:
+1 for this change, keen to see this merged ASAP.

The current master branch has a decent bug (introduced in ab5e5ff586a701e988a724e896df090eb6d1c8e0), which is causing CSS to be excluded from prod builds under the default config. If I make a fresh clone and check out master right now, then `npm run build.prod && stat dist/prod/css` will show 'file not found', and the browser gives an unstyled page.

The expected behaviour is restored by either pulling in the change in this pull request, or running `git revert ab5e5ff586a701e988a724e896df090eb6d1c8e0`.

PR #1812, opened 2017-02-12
add google analytics

Shyam-Chen:


tarlepp:
Hmm, maybe some individual gulp task should take of case of this ? eg. using https://www.npmjs.com/package/gulp-ga

mgechev:
It'll be better if the ID is configurable from project.config.ts. You can use the underscore (lodash) template syntax.

src/client/index.html
```diff
@@ -16,6 +16,15 @@
 
   <sd-app>Loading...</sd-app>
 
+  <script defer>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+    ga('create', 'UA-XXXXXXXX-X', 'auto');
```
mgechev: Use interpolation with lodash template syntax and move the ID to config.

Shyam-Chen: done

tools/config/seed.config.ts
```diff
@@ -147,6 +147,12 @@ export class SeedConfig {
   APP_TITLE = 'Welcome to angular-seed!';
 
   /**
+   * Tracking ID.
+   * @type {string}
+   */
+  Google_Analytics_ID = 'UA-XXXXXXXX-X';
```
mgechev: Let's make it uppercase to be consistent with the other configuration properties.

mgechev:
I added a comment. Let's change `Google_Analytics_ID` to `GOOGLE_ANALYTICS_ID`.

mo-ba:
pls comment it out, i think its a bad idea, i always have to delete those lines

mo-ba:
had to delete it 3 times in the last month

Shyam-Chen:
If you do not use it can be ignored directly.

mo-ba:
thats not true, it makes at least two http requests to google servers by default:
```
:authority:www.google-analytics.com
:method:GET
:path:/analytics.js
:scheme:https
accept:*/*
accept-encoding:gzip, deflate, sdch, br
accept-language:en-US,en;q=0.8,de;q=0.6
cache-control:no-cache
pragma:no-cache
referer:http://localhost:5555/
user-agent: XXXXXX
```
and 
`https://www.google-analytics.com/collect?XXXXXX`

PR #1824, opened 2017-02-18
HomeModule imports CommonModule 2 times

mpetkov:
Removed the extra CommonModule import which is already being imported by SharedModule. This is very minor but wanted to fix it.


mgechev:
Thanks for the catch!

Issue #1821, opened 2017-02-16
Test.watch not working


Cselt:
**I'm submitting a ...**  (check one with "x")
```
[x ] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see use [gitter](https://gitter.im/mgechev/angular2-seed) or [stackoverflow](https://stackoverflow.com/questions/tagged/angular2)
```

**Current behavior**
Test.watch script works only for the first time. Any change after the start will end up in an error.

**Expected behavior**
Watch files and run tests.

**Minimal reproduction of the problem with instructions**
Steps to reproduce:
1. clone repository
2. run `npm run test.watch`
3. make a change in any file
4. you will see an error message: 404 not found.

**Please tell us about your environment:**

* **Angular Seed Version:** `bdce1ef`

* **Node:** `node --version` = v6.9.4
* **OS:** Windows 10

Karasuni:
+1 - I am observing the same issue. Repeated tests during `start.deving` do work.

```
Chrome 56.0.2924 (Windows 7 0.0.0) ERROR: '(SystemJS) XHR error (404 Not Found) loading http://localhost:9876/base/dist/dev/app/home/home.component.spec.js
	Error: XHR error (404 Not Found) loading http://localhost:9876/base/dist/dev/app/home/home.component.spec.js
	    at XMLHttpRequest.wrapFn [as _onreadystatechange] (http://localhost:9876/base/node_modules/zone.js/dist/zone.js?ab03b4d15db0adaeada6236a504a0823156c8a72:1039:29) [<root>]
	    at Zone.runTask (http://localhost:9876/base/node_modules/zone.js/dist/zone.js?ab03b4d15db0adaeada6236a504a0823156c8a72:151:47) [<root> => <root>]
	    at XMLHttpRequest.ZoneTask.invoke (http://localhost:9876/base/node_modules/zone.js/dist/zone.js?ab03b4d15db0adaeada6236a504a0823156c8a72:345:33) [<root>]
	Error loading http://localhost:9876/base/dist/dev/app/home/home.component.spec.js'
```

mgechev:
@Cselt fixed the issue in https://github.com/mgechev/angular-seed/pull/1828.

PR #1828, opened 2017-02-20
fix test.watch error

Cselt:
This small fix will fix issue #1821


mgechev:
Thank you for the catch!

PR #1867, opened 2017-03-14
fix getConfig

odk211:
I really sorry, previous PR do not work...

I fix it.

Or please revert previous PR


mgechev:
It's weird that the build passed. Maybe it'll be best to revert it back because there might be code which depends on the TypeScript style exports.

PR #1878, opened 2017-03-23
Fix mispelling

idready:
Just a small fix on a mispelled word ^^


mgechev:
Thanks for the fixes!

PR #1885, opened 2017-03-31
Small fix to make possible generate not empty message file for localization

zahar-g:
I found it useful for other developers who use angular-seed to build their own apps.
I've just worked on localization for our app and can't get why it generate empty file. But this line is fixing issue. Other things from this instruction https://github.com/mgechev/angular-seed/wiki/Enabling-Angular-i18n-native are correct and very helpful.


mgechev:
Thanks for the PR!

PR #1891, opened 2017-04-04
provide rollup includes and named exports in project config

nmelv170:
We came across a few dependencies where we had to edit this seed task to include in rollup. We would rather follow the pattern of supplying the config via project.config w/out editing the seed task. Let us know what you think.

Thanks


mgechev:
Thanks for the PR!

PR #1920, opened 2017-05-05
fix(tools): cache SCSS files only in prod build

siovene:
Just like it's done for the CSS files, SCSS files should also only
cached on production build. As things were before this patch, I would
not get a rebuild of my SCSS files (as per the progeny dependency tree)
when I updated one of them. This was problematic when working in `npm
start` or `npm run start.deving` mode, and resulted in broken UI that


mgechev:
Thanks for the catch!

PR #1933, opened 2017-05-16
Add multiple language support

nhutcorp:
New gulp tasks added for generate the messages.xlf file or merge to existing translations.


nhutcorp:
@treyrich I have made some changes according to your suggestion. Please check it again and let me know if there is any part need to change.

treyrich:
Awesome LGTM, but I think that @mgechev should review as well to make sure that things are inline with the project goals. In particular I'm not sure if it would be more correct to the style of the project to break out the gulp tasks further. I'll wait for him to comment though.

mgechev:
@nhutcorp thanks for the PR! I'll take a look as soon as possible.

tools/tasks/seed/lang.build.ts
```diff
@@ -0,0 +1,7 @@
+import * as runSequence from 'run-sequence';
+import Config from '../../config';
+
+export = () => {
+    Config.BUILD_TYPE = 'prod';
```
mgechev: Would you change it to two spaces?

mgechev: Mutating the `Config` object within tasks doesn't seem like a good idea.

tools/tasks/seed/lang.merge.ts
```diff
@@ -0,0 +1,63 @@
...
+import * as gulpLoadPlugins from 'gulp-load-plugins';
+import * as runSequence from 'run-sequence';
+import { join } from 'path';
+import Config from '../../config';
+
+const plugins = <any>gulpLoadPlugins();
+const sourceElements: any[] = [];
+let langEmpty: boolean = true;
+
+gulp.task('get-source', function() {
```
mgechev: Dividing it into three different files doesn't seem like a bad idea. It'll follow the current structure and I don't think it'll be an overhead.

In the same time only a single task is being exported, so I suppose it's better to leave it the way it is.

Would you change the task names to have some special prefix to make sure we won't override custom user tasks? A suffix indicating that this is an internal task regarding the i18n seems enough (`_i18n`, `internal.i18n`, `seed.i18n`, whatever).

tools/tasks/seed/lang.merge.ts
```diff
@@ -0,0 +1,63 @@
...
+    return gulp.src(join(Config.TMP_DIR, 'messages.xlf'))
```
mgechev: Also, would you switch to two spaces?

tools/tasks/seed/lang.merge.ts
```diff
@@ -0,0 +1,63 @@
...
+gulp.task('merge-to-translations', function() {
+    let langDest = join(Config.ASSETS_SRC, 'locale');
+    return gulp.src(join(langDest, '*.xlf'))
+        .pipe(plugins.cheerio({
+            run: function($: any, file: any) {
+                let sourceIds:string[] = [];
+                langEmpty = false;
+                 for (var sourceElement of sourceElements) {
+                    let id = $(sourceElement).attr('id');
+                     sourceIds.push(id);
```
mgechev: Would you fix the indentation here?

tools/tasks/seed/lang.merge.ts
```diff
@@ -0,0 +1,63 @@
...
+                    let targetElement = $('#' + id);
+                     if (targetElement.length === 0) {
+                         // missing translation
+                         $('body').append(sourceElement);
+                     }
+                 }
+                 $('trans-unit').map((function() {
+                    let id = $(this).attr('id');
+                    let existing = sourceIds.find((item) => { return item === id; });
```
mgechev: `sourceIds.find((item) => { return item === id; });` -> `sourceIds.find((item) => item === id);

mgechev:
@nhutcorp dropped some minor comments.

nhutcorp:
@mgechev Thank you for your detailed checking, I will try to group the tasks and fix code format soon.

README.md
```diff
@@ -102,6 +102,26 @@ Your project will be compiled ahead of time (AOT), and then the resulting bundle
...
 - During a production build, CommonJs modules will be automatically converted to ES6 modules. This means you can use them and/or require dependencies that use them without any issues.
 
+# Internationalization
+
+Put `i18n` attribute to your html tag to mark it for translation, more information here: https://angular.io/docs/ts/latest/cookbook/i18n.html
+
+## Create a translation source file
+
+```bash
+# Your translation file will be generated here `src/client/assets/locale`
```
Hirse: Could this be made configurable?
I would personally prefer not to have generated files in the `src`.

Also depending on your translation process, it would be good to keep the source and translated files separate.

nhutcorp:
@Hirse That's good idea, I'll try that way. @mgechev Can I add LOCALE_PATH variable to `seed.config.ts` file and modify code of `compile.ahead.prod.ts` file?

mgechev:
Sure, I will wait before merge.

README.md
```diff
@@ -102,6 +102,24 @@ Your project will be compiled ahead of time (AOT), and then the resulting bundle
...
+
+```bash
+# Your translation file will be generated here `dist/locale`
+$ npm run i18n
+```
+
+## Production build with your language
+
+```bash
+# Build prod app with the language file `dist/locale/messages.en.xlf`
```
Hirse: Sorry, I didn't explain exactly.

I meant to have the generated files be in `dist` for example, but have the translated files in `src`.
The translation process might be something like:
1. Developer adds `i18n` to code
2. Generate `dist/locale/messages.xlf`
3. Give `messages.xlf` to translators
4. Receive `messages.en.xlf`, `messages.ja.xlf`, ... from translators
5. Check those in version control in `src/assests/locale` for example

Does that sound reasonable to you?

nhutcorp: Did you mean we divide the translation task to 2 process:
First, run `npm run i18n` to generate `dist/locale/messages.xlf` file.
(Developer will translate this file and put it into `dist/e2e` folder)
Second, run `npm run i18n.merge` (for example) to merge to existing translations in `src/assests/locale`

PR #1939, opened 2017-05-20
Update i18n message extraction

donvaughn:
This PR fixes the i18n message extraction by:
1. adding a `--i18nFormat=xlf` parameter to the `i18n` script entry in `package.json`
1. adding the `src/tsconfig.i18n.json`, which:
    - solves the [ngfactory](https://github.com/mgechev/angular-seed/issues/1582) error
    - allows the output directory to be changed, via the `genDir` parameter
1. adds the missing `tools/tasks/seed/clean.i18.ts` file, which removes the `tsconfig.i18n.json` file that `ng-xi18n` moves into the dist/dev folder and also solves [this issue](https://github.com/mgechev/angular-seed/issues/1729)

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "e2e": "protractor",
     "e2e.live": "protractor --elementExplorer",
     "gulp": "gulp",
-    "i18n": "ng-xi18n -p src/client/tsconfig.json && gulp clean.i18n",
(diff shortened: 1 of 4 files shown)


mgechev:
Thanks for the PR. Would you resolve the merge conflict?

donvaughn:
I see that [PR 1933](https://github.com/mgechev/angular-seed/pull/1933) solves the same issues after adding the `i18nFormat` param. I'll close this one.

PR #1974, opened 2017-06-14
adicionando pagina de teste

felipecamargo:
comentarios


mgechev:
Hey @felipecamargo, I don't think we should extend the base application.

PR #2004, opened 2017-07-07
Update package.json

JunusErgin:
vinyl was missing in dev dependencies, npm start was not working after npm install with a clean project.


mgechev:
@JunusErgin thanks for the PR! It seems `vinyl` is dependency of `vinyl-fs`, which is dependency of `gulp`. What error did you get?

JunusErgin:
Hey Minko, i don't remember the exact error message. `npm install` didn't run through and told me that vinyl was missing.

PR #2005, opened 2017-07-07
Sync local changes with docker container

maxklenk:
The current docker development configuration only copies the users code into the container in the build process. New changes to the code are not copied and therefore the watchers and livereload does not work.

This PR mounts the project directory on the host to the container, allowing you to modify the code without having to rebuild the image. It does not sync the `node_modules` folder to avoid conflicts of OS specific packages and bindings.


mgechev:
Thanks for the PR!

PR #2017, opened 2017-07-13
Revert zone js

pbazurin-softheme:
Temporary fix for https://github.com/mgechev/angular-seed/issues/2016


mgechev:
Would you provide source?

pbazurin-softheme:
Of the problem?

pbazurin-softheme:
Found a similar issue - https://github.com/angular/zone.js/issues/832.
And tested locally.
Unfortunately don't know about details. Hope it would be fixed in the next zonejs version.

PR #2025, opened 2017-07-18
fix: css asset ordering

JimbeanZN:
include APP_ASSETS before building css of seed.

This will allow people using libraries like Bootstrap to not have their custom styles overridden. With the previous ordering any changes made in you main.css/main.scss files would be overridden.


JimbeanZN:
fix issue with spacing.

PR #2059, opened 2017-08-18
add version to envConfig

Karasuni:
Expose package version to the application


Karasuni:
On the move so I edited the files from web. Are you able to squash the commits before merge or should I update the PR itself?

PR #2078, opened 2017-09-14
Add translation file checking

nhutcorp:
Check translation file exist when build with --lang parameter


mgechev:
Nice fix! Thanks @nhutcorp!

Issue #2057, opened 2017-08-17
Add HttpClientModule to seed config


aGoncharuks:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
[ ] support request => Please do not submit support request here, instead see use [gitter](https://gitter.im/mgechev/angular2-seed) or [stackoverflow](https://stackoverflow.com/questions/tagged/angular2)
```

**Current behavior**

No HttpClientModule support

**Expected behavior**

Support HttpClientModule

**Minimal reproduction of the problem with instructions**
New Http client was added in angular 4.3.0 providing several useful features:
* Typed, synchronous response body access, including support for JSON body types
* JSON is an assumed default and no longer needs to be explicitly parsed
* Interceptors allow middleware logic to be inserted into the pipeline
* Immutable request/response objects
* Progress events for both request upload and response download
* Post-request verification & flush based testing framework

Is it possible to include HttpClientModule in seed config? I know it shouldn't be a big deal to add it to SystemJS config, but for some reason I failed to do it and none of solutions offered in #2046 and #2033 helped. I think it would be great to have it in seed by defaul. It also will be very helpful to everyone who will need to import this module in their existing project based on angular seed and might experience similar problems as we did.

Feature in angular repo: https://github.com/angular/angular/commit/37797e2
Docs: https://angular.io/guide/http

mgechev:
Closed by https://github.com/mgechev/angular-seed/pull/2083

PR #2083, opened 2017-09-21
Added dependencies for Angular's  http client

tobiaseisenschenk:
Angular's http client from @ angular/common/http should be a dependency in the seed config, not in the project config as pointed out in #2057.


mgechev:
Did you test if the production build works with this configuration?

tobiaseisenschenk:
Yes, tested and works!

Issue #2084, opened 2017-09-22
httpclientmodule error 


gabhub1:
**I'm submitting a ...**  (check one with "x")
```
[x ] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => Please do not submit support request here, instead see use [gitter](https://gitter.im/mgechev/angular2-seed) or [stackoverflow](https://stackoverflow.com/questions/tagged/angular2)
```

**Current behavior**

After importing HttpClientModule from - 
import { HttpClientModule } from '@ angular / common/http';   ( i had to put spaces because github collapsed text for some reason )
error pops up
Error trace:
localhost/:53 SyntaxError: Unexpected token <
    at eval (<anonymous>)
    at evaluate (http://localhost:5555/node_modules/systemjs/dist/system.src.js?1506079427789:2821:16)
    at http://localhost:5555/node_modules/systemjs/dist/system.src.js?1506079427789:3620:21
    at dynamicExecute (http://localhost:5555/node_modules/systemjs/dist/system.src.js?1506079427789:1144:26)
    at doEvaluate (http://localhost:5555/node_modules/systemjs/dist/system.src.js?1506079427789:1091:13)
    at ensureEvaluate (http://localhost:5555/node_modules/systemjs/dist/system.src.js?1506079427789:999:13)
    at http://localhost:5555/node_modules/systemjs/dist/system.src.js?1506079427789:1017:20
    at doEvaluate (http://localhost:5555/node_modules/systemjs/dist/system.src.js?1506079427789:1089:11)
    at ensureEvaluate (http://localhost:5555/node_modules/systemjs/dist/system.src.js?1506079427789:999:13)
    at http://localhost:5555/node_modules/systemjs/dist/system.src.js?1506079427789:1017:20 Not expecting this error? Report it at https://github.com/mgechev/angular-seed/issues

**Expected behavior**

**Minimal reproduction of the problem with instructions**

download angular-seed, install, import httpclient and run  with npm start

* **Angular Seed Version:** 

Merge pull request `eace08d` , Sep 21, 2017

mgechev:
@tobiaseisenschenk would you take a look at this?

ahmadqarshi:
To fix this problem add the entry for `tslib` into `additionalPackages`  inside "tools/config/project.config.ts" file like below:

    let additionalPackages: ExtendPackages[] = [
      {
        name: 'tslib',
        path: 'node_modules/tslib',
        packageMeta: {
[...]

PR #2091, opened 2017-09-28
Fixed HttpClientModule import issue #2084

ahmadqarshi:
Fixed "SyntaxError: Unexpected token <" issue [#2084](https://github.com/mgechev/angular-seed/issues/2084) by adding the tslib into additionalPackages inside "tools/config/seed.config.ts" .

### DIFF
--- a/tools/config/seed.config.ts
+++ b/tools/config/seed.config.ts
@@ -357,8 +357,8 @@ export class SeedConfig {
   ENABLE_SCSS = ['true', '1'].indexOf(
     `${process.env.ENABLE_SCSS}`.toLowerCase()
   ) !== -1 ||
-  argv['scss'] ||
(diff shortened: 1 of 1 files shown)


mgechev:
Thanks for the PR!

brian428:
Correct me if I'm wrong, but wouldn't this also require adding `"importHelpers": true` to `/src/client/tsconfig.json`?

mgechev:
Eventually the bundles might be smaller if we require the helpers manually. Need to be tested for verification.

brian428:
FYI, I just tried it and it shaved about 3% off the size of my `app.js`. Not a huge difference but it is a bit smaller.

mgechev:
Thanks for giving it a try!

PR #2099, opened 2017-10-13
Feature/stylelint

fr-esco:


mgechev:
@fr-esco thanks for the PR. For feature PRs I'd propose to first open an issue in order to discuss if we want to extend the functionality of the seed.

I'd prefer to not include additional tasks because they require extra maintenance and it's not clear how many users would want to use them. Lets keep the PR open for a bit and collect opinions.

fr-esco:
There was already a `css-lint` task mentioned in README, so I guessed it'd be useful to have it actually implemented.

I agree an Issue would be a better place for discussions. I'll open one next time for sure. Thanks.

mgechev:
@fr-esco thanks for pointing this out! Dropped the `css-lint` task from the READMEs.

mgechev:
Lets not introduce additional dependencies for now.

PR #2100, opened 2017-10-13
Feature/core module

fr-esco:


mgechev:
Lets keep the application as simple as possible.

PR #2107, opened 2017-10-20
Simplify docker-compose.production using multi-stage build

ArnaudPel:
Use the multi-stage build feature of Docker 17.05+ do improve production build
https://docs.docker.com/engine/userguide/eng-image/multistage-build/


mgechev:
Thanks for the PR!

PR #2131, opened 2017-11-15
TS Lint rules update

vyakymenko:
Since `tslint` is growing very fast, I decided to upgrade `tslint` rules and adapt code for them.


vyakymenko:
Interesting issue with `AppVeyour`, some problems with the config I think.

vyakymenko:
Hi @mgechev, I don't think that this will be very restrictive because most of this rules are used in `@angular/cli`.

PR #2142, opened 2017-11-20
SCSS bug refactor

vyakymenko:
Update:
* Small task refactoring, remove `gulp-if` to minimize `devDependencies` because forgot that we have `gulp util` that provide `util.nope()`.


vyakymenko:
Hi @mgechev , I modified hotfix for `SCSS` files, let's use this solution instead of adding `gulp-if` dependency,

Thanks!

mgechev:
@vyakymenko even better

vyakymenko:
My mistake, forgot that we use ‘gulp util’ :D

PR #2210, opened 2018-02-09
Cypress integration

### LINKED ISSUE #2196: Switch to Cypress for e2e testing
Consider switching to Cypress to improve the development experience and prevent random build failures because of glitches in the webdriver.

I believe the TypeScript support of Cypress is mature enough for this step. @bahmutov, any thoughts?

salemdar:
Closes #2196


salemdar:
CI fails because no server starts. I think there's no official gulp integration for cypress. We can use `npm-run-all` like in [this example](https://github.com/cypress-io/cypress-example-todomvc/blob/master/package.json). What do you think @mgechev ?

mgechev:
Can we go with `npm start & cypress run`? I am fine with `npm-run-all` too, if we cannot get away from this dependency.

salemdar:
Maybe taking `gulp-exec` route would be more elegant. We can start server, and run the tests easily imho.

mgechev:
❤️

Issue #2168, opened 2017-12-14
Create CoreModule


salemdar:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
[ ] support request => Please do not submit support request here, instead see use [gitter](https://gitter.im/mgechev/angular2-seed) or [stackoverflow](https://stackoverflow.com/questions/tagged/angular2)
```

**What is the motivation / use case for changing the behavior?**
[Angular Style Guide](https://angular.io/guide/styleguide#core-feature-module) suggests creating a `CoreModule` and collecting application wide classes such as `NavbarComponent` in there. Since this repo follows the guide, I think we can create one and move `NavbarComponent` there.

vyakymenko:
Hello @salemdar, I think that if we will create `core.module`, this will be for `ToolbarComponent` and `NavbarComponent` not only `NavbarComponent`.

PR #2220, opened 2018-02-26
Create core module

vyakymenko:
Add Core module.
_Note:_ Per issue #2168.


lauruka:
Shouldn't navbar and toolbar be in Shared folder instead of Core?

PR #2240, opened 2018-04-06
Upgrade dependencies (especially Jasmin-core)

salemdar:
https://github.com/angular/zone.js/issues/1015


mgechev:
Thanks for the update!

PR #2241, opened 2018-04-07
Use RxJS pipeables instead of lettables

salemdar:
https://github.com/ReactiveX/rxjs/blob/master/doc/pipeable-operators.md


src/client/app/operators.ts
```diff
@@ -1,7 +1,7 @@
-// rxjs
-import 'rxjs/add/observable/throw';
-import 'rxjs/add/operator/map';
-import 'rxjs/add/operator/catch';
+// rxjs: Add only if lettable operators are being used. Pipeable operators don't require this.
```
mgechev: Would you drop this file completely? We don't want to encourage people to write imports with side effects.

mgechev:
Left one comment only. LGTM otherwise!

src/client/app/operators.ts
salemdar: Sure.

Btw, When we use `import { catchError } from 'rxjs/operators/catchError';`, it gets bundled just fine. But when we use `import { catchError } from 'rxjs/operators';`, we get all the operations in the prod bundle. There's an easy webpack solution for that, but do you think it is possible with SystemJS?

src/client/app/shared/name-list/name-list.service.ts
```diff
@@ -1,7 +1,9 @@
 import { Injectable } from '@angular/core';
 import { HttpClient } from '@angular/common/http';
 import { Observable } from 'rxjs/Observable';
-// import 'rxjs/add/operator/do';  // for debugging
+import { ErrorObservable } from 'rxjs/observable/ErrorObservable';
```
mgechev: Once version 6 is out we should update the imports as well.

PR #2248, opened 2018-04-17
Docker composer hotfix for cypress

vyakymenko:
Without a copy of `cypress` directory we can't create development docker container, because `$ npm install` required `cypress` directory.


mgechev:
👍

PR #2271, opened 2018-05-12
add bundle rxjs for less request

aelbore:
Added task to bundle rxjs for less request.
currently there are 271 request when you npm start or start the app.
with this bundle it would be 55 request only.


aelbore:
update the karma config and test config

Issue #2288, opened 2018-06-11
Remove build.prod and build.prod.aot


vyakymenko:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[X ] feature request
[ ] support request => Please do not submit support request here, instead see use [gitter](https://gitter.im/mgechev/angular2-seed) or [stackoverflow](https://stackoverflow.com/questions/tagged/angular2)
```

**Current behavior**
Since we have a lot issues with `build.prod` or `build.prod.aot`, for [example](https://github.com/mgechev/angular-seed/issues/2277).

I think it's a time to remove `build.prod` and `build.prod.aot` since all developers use `build.prod.rollup.aot` and I think that only last task make sense, we also can rename it to `build.prod` and remove other options and drop a lot of dependencies and optimize tasks section.

**Expected behavior**
To avoid problems with supporting all 3 build tasks `build.prod`, `build.prod.aot` and `build.prod.rollup.aot` I propose to use only `build.prod.rollup.aot` and rename it to `build.prod`. Also, refactor a lot of dependencies and tasks.

P.S., In my opinion, there is no sense to provide for developers unoptimized build technics and confuse them. Let's provide maximum out of the box.

**Minimal reproduction of the problem with instructions**
For example `build.prod` is not working while `build.prod.rollup.aot` work like a charm. And there is no reason to work twice.

**What is the motivation / use case for changing the behavior?**
Knowledge sharing must be efficient, we need to share knowledge that provides the maximum for developers and one day someone will have an idea based on that knowing how to improve it more and more. We need to move on.

**Please tell us about your environment:**
All OS, All IDE.

mgechev:
This makes sense. Temporary we can invoke rollup for all these tasks and print a warning message for some of them that they are deprecated.

vyakymenko:
I hope that I'll have some time very soon to take a look at this issue.

pvskasyrskiy:
@vyakymenko, @mgechev Guys please do not remove build.prod.aot build type in project. I have project which based on angular-seed but can not use build.prod.rollup.aot due issue with Ag-grid https://github.com/ag-grid/ag-grid-angular/issues/112

PR #2290, opened 2018-06-13
Exclude Cypress and Karma from docker images

ArnaudPel:
Currently, the test code installed in the docker containers cannot be used:
Eg. to run Karma, the command `npm test` would have to be run from inside the container. On top of not being documented in the readme, this would lead to major issues, eg. the absence of a browser. Generally speaking, AFAIK, handling GUI apps in Docker is far from trivial.

Therefore, there is no point in wasting time and bandwidth waiting for Cypress to install when building the images.

If this change is accepted, may I ask someone to help and have a look at what other dependencies could go into `optionalDependencies` ? Basically, every dependency only required in the test env can go there.

Thanks


ArnaudPel:
Incidentally, this commit also fixes the production docker image build.
Until #2288 is fixed.

mgechev:
The buld is still failing. Let's fix it before we merge.

ArnaudPel:
Yes I'm a bit lost as to why Travis won't install Cypress, but seems to find Karma without any issue. I put both in the same `optionalDepedencies` section.

As can be seen in the following commits, I tried the following changes:
- don't put `cypress/` and `cypress.json` in `.dockerignore`  -- even though I see no mention of our docker config in `.travis.yml`
- move Cypress back to `devDependencies`

None of the above work. In want for something else to try, I'm tempted to revert all the changes on this branch to check that the Cypress install issue is indeed coming from my change.

The Appveyor build has all tests (Karma, Cypress) passing, but crashes right after Cypress completes :(

ArnaudPel:
Ok I got yet another idea, try to revert `package-lock.json` to master.

ArnaudPel:
Ok so that's what I was suspecting: even after reverting everything to master, I still get the same error:
`No version of Cypress is installed in: /home/travis/.cache/Cypress/3.0.1/Cypress`

I've never worked with Travis or any Github CI before, so I can't invest more effort into that sorry. All I can say is that the Appveyor build runs all tests successfully, even though it is not aware of it :)

ArnaudPel:
Trying again in #2309

PR #2302, opened 2018-06-25
add resolutions field to package.json

salemdar:
We should not forget to remove `resolutions` block from `package.json` once Angular updates to TypeScript >2.8.


salemdar:
About appveyor: https://github.com/cypress-io/cypress/issues/1841

PR #2309, opened 2018-07-02
Exclude Cypress and Karma from docker images

ArnaudPel:
Trying #2290 again on a new branch, because the builds did not seem to be reset properly between commits on the other branch


ArnaudPel:
Seems to work !

package.json
```diff
@@ -50,11 +50,23 @@
     "node": ">=6.9.0",
     "npm": ">=3.10.3"
   },
+  "optionalDependencies": {
+    "@bahmutov/add-typescript-to-cypress": "^2.0.0",
```
mgechev: Not sure about that. There's no standard property called `optionalDependencies` in `package.json`.

ArnaudPel: Is this not standard ? https://docs.npmjs.com/files/package.json#optionaldependencies

mgechev: My bad. Let's merge.

ArnaudPel: In my opinion the issue with our `package.json` is that a lot of things that are currently listed as `devDependencies` are actually needed to build the prod environment as well: `gulp`, `rollup`, and `types`, for example, are needed by the build process.

In this regard, I don't quite see the point of the current split between `dependencies` and `devDependencies`. AFAICT there's nowhere we can run `npm install --production`.

Since I didn't want to bring that issue up, I tried instead to look for something like `testDependencies`, but the closest I could find was this `optionDepedencies`.

To sum up, I currently see two options only:
- move all dependencies required by the prod build to `dependencies`
- keep change as is

ArnaudPel: Ah ok you accepted already :)

PR #2312, opened 2018-07-05
Fix test.watch #2305

ArminZol:
Currently, test.watch does not work on the starter repo. 

This will allow it to run without the `completion callback called too many times` error referenced in #2305


ArminZol:
Karma.conf file is same as the one on the master branch. Shouldn't show that there is a change.

karma.conf.js
```diff
@@ -28,10 +28,10 @@ module.exports = function(config) {
       // Zone.js dependencies
       'node_modules/zone.js/dist/zone.js',
       'node_modules/zone.js/dist/long-stack-trace-zone.js',
+      'node_modules/zone.js/dist/proxy.js',
```
mgechev: Is this required?

ArminZol: My previous commit fixed the async test cases. This is the same file that is on the master branch. Project was rebased as well but for some reason it still shows it as a change.

